PR #1, opened 2021-11-23
Doc 956 `/current-service-status` endpoint

lennessyy:
The callouts `<1>, <2>, <3>` will not look distracting when rendered (can see an example here: https://docs-tigergraph.netlify.app/tigergraph-server/3.3/gsql-shell/gsql-shell#_multi_line_commands)


chengjie-qin (approved):
lgtm

PR #3, opened 2021-12-03
Doc 958 sso

lennessyy:
SSO doc update

### DIFF
--- a/modules/user-access/pages/sso.adoc
+++ b/modules/user-access/pages/sso.adoc
@@ -1,5 +1,6 @@
 = Single Sign-On
 :experimental:
+:sectnums:
 
 The Single Sign-On (SSO) feature in TigerGraph enables you to use your organization's identity provider (IDP) to authenticate users to access TigerGraph GraphStudio and Admin Portal UI.
 
@@ -22,6 +23,7 @@ We assume you already have TigerGraph up and running , and you can access GraphS
 If you enabled SSL connection, change `http` to `https`. If you changed the Nginx port of the TigerGraph system, replace 14240 with the port you have set.
 
 == *Configure Identity Provider*
+:sectnums!:
 
 Below are detailed instructions for identity providers that we have verified.
 Please consult your IT or security department for how to configure the identity provider for your organization if it is not listed here.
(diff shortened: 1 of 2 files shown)


modules/user-access/pages/sso.adoc
```diff
@@ -1,28 +1,35 @@
...
 For supporting additional IDPs, please inquire sales@tigergraph.com and submit a feature request.
 
-In order to use Single Sign-On , you need to perform four steps :
+In order to use Single Sign-On , you need to perform three steps :
 
 . Configure your identity provider to create a TigerGraph application.
-. Provide information from your identity provider to enable TigerGraph Single Sign-On .
+. Provide information from your identity provider to enable TigerGraph Single Sign-On.
 . Create user groups with proxy rules to authorize Single Sign-On users.
-. Change the password of the tigergraph user to be other than the default, if you haven't done so already.
```
songrenchu, line 17: This is still needed, otherwise GraphStudio will login as tigergraph by default

modules/user-access/pages/sso.adoc
```diff
@@ -1,28 +1,35 @@
...
 
-We assume you already have TigerGraph up and running , and you can access GraphStudio UI through a web browser using the URL: http://tigergraph-machine-honestname:14240
+== Before you begin
+* Ensure that you can access GraphStudio UI through a web browser using the URL: http://tigergraph-machine-honestname:14240
+** If you enabled SSL connection, change `http` to `https`.
+If you changed the Nginx port of the TigerGraph system, replace 14240 with the port you have set.
+* Ensure that you have xref:enabling-user-authentication.adoc[enabled GSQL authentication].
+If you don't enable authentication, SSO configuration will not work and any user will be logged in automatically.
```
songrenchu, line 25: Why is enabling authentication needed?

lennessyy: In our enabling authentication how-to, all you do to enable GSQL authentication is to change the password of the default user. That's why I moved that step to this prerequisite. I can make this clearer by adding an anchor to the page and say

> Ensure that you have enabled GSQL authentication by changing the password of the default user.

modules/user-access/pages/sso.adoc
lennessyy: I listed this as a prerequisite instead. I think if a user is trying to do SSO, it should be expected that they should have enabled authentication.

songrenchu (approved):

PR #8, opened 2021-12-20
Memory mgmt doc 688

lennessyy:
- Added page for memory management
- Renamed gadmin module to system management


lennessyy:
@elliotmartinTG Can you take a look at this draft when you get a chance?

elliotmartinTG (approved):
LGTM. 2 nitpicks: 
query memory usage in GPE logs will only be there for GPR mode queries
Mentions that SysAlertFreePct and SysMinFreePct are configured via env vars, but no mention of how. Maybe this is on purpose though?

lennessyy:
- Added the procedure to configure the thresholds
- @elliotmartinTG Thanks for noting that query memory usage in GPE logs only works for GPR queries. Any way to monitor UDF query's memory usage?

PR #9, opened 2021-12-23
Doc 968 built in endpoints

alexthomasTG:


modules/API/pages/built-in-endpoints.adoc
```diff
@@ -1625,7 +1664,7 @@ $ curl -X POST --data-binary @add_id6.json \
   "code": "REST-0003"
 }
 ----
```
lennessyy, line 1666: When you use tabs, a tab is in an open block indicated by `--`. When you end one it needs to be closed. This one doesn't seem to be closed.

lennessyy (changes requested):
Great job fixing the formats and discovering that we don't always need an open block for tabs to work.  This will save us time when doing more of this kind of formatting. 

However, in instances where you did use open blocks, please update and close them.

lennessyy:
@a-m-thomas Did you test if the tabs format you are using compiles in Antora? I just tried to run a build from this branch, and it didn't seem to compile, and it seems to be because of the tab format: 

```
[10:37:40.281] FATAL (antora): details.getBlocks is not a function
    Cause: TypeError
        at /Users/lenny.chen/workstation/doc-site/libs/tabs-block.js:54:28
        at Array.forEach (<anonymous>)
        at constructor.<anonymous> (/Users/lenny.chen/workstation/doc-site/libs/tabs-block.js:49:27)
        at Object.apply (/Users/lenny.chen/workstation/doc-site/node_modules/@asciidoctor/core/dist/node/ascii
```

modules/API/pages/built-in-endpoints.adoc
```diff
@@ -2364,18 +2504,24 @@ To run query `hello` on a graph named `social`, and the query parameter is of ty
 .Running a query via HTTP request
 
 [source,bash]
-----
```
lennessyy, line 2367: This part looks a bit weird. Are you sure this will render the way you expect?

modules/API/pages/built-in-endpoints.adoc
```diff
@@ -2438,23 +2584,28 @@ For example, if you have a cluster of three nodes, each with 8 vCPUs, then a que
...
-[source,bash]
-----
-curl -X POST -d '{"p":{"id":"Tom","type":"person"}}' \
-"http://localhost:9000/query/social/hello"
-
-# Specify that the query run on the first replica
-curl -X POST -d -H "GSQL-REPLICA: 2" '{"p":{"id":"Tom","type":"person"}}' \
+[source.wrap,bash]
+--
+curl -X POST -d '{"p":{"id":"Tom","type":"person"}}'
```
lennessyy, line 2589: Still open blocks(--) for code blocks (----). Have you verified that this renders correctly?

modules/API/pages/built-in-endpoints.adoc
```diff
@@ -2579,7 +2750,7 @@ curl -X GET "localhost:9000/abortquery/poc_graph?requestid=16842763.RESTPP_1_1.1
 
 | `url`
 | No
-| The endpoint whose running queries to abort. You must specify the base of the endpoint's URL, but then use a wildcard to allow for different parameters. For example, to abort all running queries for the endpoint `/sleepgpe`, use `url=/sleepgpe.*`
+| The endpoint on which to abort running queries. You must specify the base of the endpoint's URL, but then use a wildcard to allow for different parameters. For example, to abort all running queries for the endpoint `/sleepgpe`, use `url=/sleepgpe.*`
```
lennessyy, line 2753: I think this rephrasing is not correct. 

The original meaning is that if you have a query (NOT necessarily a GSQL query - this is confusing and we should find a better way of saying this. Query here basically refers to a request) that is running on RESTPP, you can abort it by specifying the URL of the request. So the query here belongs to the endpoint - and that possessive relationship is important.

lennessyy (changes requested):

modules/API/pages/built-in-endpoints.adoc
```diff
@@ -250,7 +264,7 @@ $ curl -X GET
 ]
 ----
 
-In the below example, the request asks for the 10 latest data points regarding memory usage:
+In the below example, `what` and `latest` are used. The request asks for the `10` latest data points regarding ``mem``ory usage:
```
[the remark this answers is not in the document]
a-m-thomas: I have no idea why the ``mem`` thing appeared...

modules/API/pages/built-in-endpoints.adoc
a-m-thomas: I thought for quite a while but couldn't come up with anything yet to improve the original, so reverted.

modules/API/pages/built-in-endpoints.adoc
```diff
@@ -641,18 +654,31 @@ If none is provided, the check will be performed on all vertex types.
...
 results. Here is a list of accepted values and their corresponding level
-of detail:`0` (default) : Only return whether the check passed and the
-list of unsynced vertex IDs`1`: In addition to the previous level, also
-return vertex count information `2`: In addition to the previous level,
-return vertex count information for every segment`4`: In addition to the
-previous level, also return the IDs of deleted vertices for every
+of detail:
+
+`0` (default) : Only return whether the check passed and the
+list of unsynced vertex IDs
```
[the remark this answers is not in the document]
lennessyy: Hi @a-m-thomas , I see that you marked this as resolved, but it doesn't look like this is a list. I was hoping that each of the option here is an item of an unordered list.

modules/API/pages/built-in-endpoints.adoc
```diff
@@ -1297,9 +1376,9 @@ This endpoint upserts vertices and/or edges into a graph. To upsert means that i
 
 ==== Atomic upsert transaction
 
-By default, the `+POST /graph/{graph_name}+` endpoint is not atomic. If something goes wrong during the process of the request,  the request data can be partially consumed by the database.
+By default, the `+POST /graph/{graph_name}+` endpoint is not atomic. If something goes wrong during the server processing stage of the request, the request data can be partially consumed by the database.
```
lennessyy, line 1379: Let's not narrow down the stage when something could go wrong. Something could go wrong during any stage of the request, making the request, server processing the request, anything. I would just say: 

"If anything goes wrong during the lifecycle of a request, the request data can be partially consumed by the database."

modules/API/pages/built-in-endpoints.adoc
```diff
@@ -2579,7 +2797,7 @@ curl -X GET "localhost:9000/abortquery/poc_graph?requestid=16842763.RESTPP_1_1.1
 
 | `url`
 | No
-| The endpoint whose running queries to abort. You must specify the base of the endpoint's URL, but then use a wildcard to allow for different parameters. For example, to abort all running queries for the endpoint `/sleepgpe`, use `url=/sleepgpe.*`
+| The endpoint whose running queries to abord. You must specify the base of the endpoint's URL, but then use a wildcard to allow for different parameters. For example, to abort all running queries for the endpoint `/sleepgpe`, use `url=/sleepgpe.*`
```
lennessyy, line 2800: typo here

lennessyy (changes requested):

modules/API/pages/built-in-endpoints.adoc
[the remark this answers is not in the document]
a-m-thomas: Oh, got it!

lennessyy (approved):

PR #13, opened 2022-01-05
[DOC-1001] fix(authentication): update docs for /requesttoken API;

JunruZhu:


modules/API/pages/built-in-endpoints.adoc
```diff
@@ -835,39 +845,44 @@ This endpoint takes a token and its associated secret and refreshes the lifetime
...
-  "message": "Refresh token successfully.",
+  "message": "Drop token successfully.",
   "token": "0mq98l9pderkaivndf820gudg923p3l0"
 }
 ----
 
-[CAUTION]
-====
-*Known bug*: The output shows the lifetime instead of the expiration time.
-====
```
JunruZhu, line 863: After some trials, it seems that the output shows the expiration time although it is a little bit weird to me. 
If I refresh the same token in sequence, I can get the following output. 

Without lifetime -> "expiration":1643995777
"lifetime": 15 -> "expiration":1641403829
"lifetime": 15 -> "expiration":1641403837
"lifetime": 15 -> "expiration":1641403883
"lifetime": 15 -> "expiration":1641403902
"lifetime": 15 -> "expiration":1641404037
"lifetime": 15000 -> "expiration":1641419046
"lifetime": 150000 -> "expiration":1641554169

From the output I get, I guess it might be a precision error.

modules/API/pages/built-in-endpoints.adoc
```diff
@@ -754,17 +760,25 @@ If authentication is enabled on RESTPP endpoints, a token needs to be included i
 
 [source,bash]
 ----
-curl -X GET "http://localhost:9000/requesttoken?secret=jiokmfqqfu2f95qs6ug85o89rpkneib3&lifetime=1000000"
+curl -X POST http://localhost:9000/requesttoken \
+     -d '{"secret": "your_secret_value", "lifetime": "100000"}' 
+
+curl --user example_username:example_password \
+     -X GET http://localhost:9000/requesttoken \
+     -d '{"graph": "example_graph"}'
```
junhyungshinAtTigergraph, line 768: First of all, I think we can keep using the arbitrary secret `jiokmfqqfu2f95qs6ug85o89rpkneib3` for better understanding, so as the example of the `PUT` and `DELETE` methods.

On the other hand, it would be better if we unify the reference of a keyword in the same context throughout the docs. For example, we use `example_username` here, but some examples in this page use `tigergraph`, and some use `<my_username>`. What do you think @lennessyy, and what would be the best practice in your opinion?

modules/API/pages/built-in-endpoints.adoc
```diff
@@ -741,9 +741,15 @@ $ curl -X GET 'http://localhost:9000/deleted_vertex_check?threadnum=10&verbose=0
 
 The endpoints in this subsection allow users to create, refresh and delete authentication tokens for requests made to the REST{pp} server. *These endpoints only exist when* xref:user-access:enabling-user-authentication.adoc[*user authentication is enabled*] *on RESTPP endpoints.*
 
-=== Request a token (`GET`)
+[CAUTION]
+====
+Please be aware to pass parameters in the json payload for `/requesttoken` api from 3.5. If the old API version is used, it could cause information leakage.
```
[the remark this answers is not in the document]
JunruZhu: How about this?

```
[CAUTION]
====
Please be aware to pass parameters in the json payload for `/requesttoken` api from 3.5. If the API version before version 3.5 is used, it could cause information leakage.
====
```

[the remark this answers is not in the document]
lennessyy: Hmm, from what I understand, this is to warn people not to use the `/requesttoken` API from 3.4 and prior? In that case, the warning perhaps should be in the older versions instead of here. If a user is reading the 3.5 documentation, we assume that they are using 3.5. 

Is the risk for information leakage high? This sounds a little scary if we say it this way, so I want to make sure we are communicating it appropriately.

modules/API/pages/built-in-endpoints.adoc
lennessyy: > First of all, I think we can keep using the arbitrary secret `jiokmfqqfu2f95qs6ug85o89rpkneib3` for better understanding, so as the example of the `PUT` and `DELETE` methods.
> 
> On the other hand, it would be better if we unify the reference of a keyword in the same context throughout the docs. For example, we use `example_username` here, but some examples in this page use `tigergraph`, and some use `<my_username>`. What do you think @lennessyy, and what would be the best practice in your opinion?

Ah, I realized that my comment didn't send yesterday.

I agree that the arbitrary password is better for helping users understand what a request looks like.

For placeholders, yes it would be nice to have consistency. I've been sticking with `example_something` unless there is a good reason not to. I think we can continue to use `example_something` and I'll formalize it in the style guide.

lennessyy (approved):

modules/API/pages/built-in-endpoints.adoc
```diff
@@ -739,148 +739,167 @@ $ curl -X GET 'http://localhost:9000/deleted_vertex_check?threadnum=10&verbose=0
 
 == Authentication
 
-The endpoints in this subsection allow users to create, refresh and delete authentication tokens for requests made to the REST{pp} server. *These endpoints only exist when* xref:user-access:enabling-user-authentication.adoc[*user authentication is enabled*] *on RESTPP endpoints.*
+The endpoints in this subsection allow users to create, refresh and delete authentication tokens for requests made to the REST{pp} server.
+*These endpoints only exist when* xref:user-access:enabling-user-authentication.adoc[*user authentication is enabled*] *on RESTPP endpoints.*
```
JunruZhu, line 744: Just take a note here. If we keep the current implementation, we need to add a warning here.

```
[CAUTION]
====
Please be aware to pass parameters in the json payload for `/requesttoken` api from 3.5. If the old API version is used, it could cause information leakage.
====
```

lennessyy: Hi Junru, I think I will add this in the release note instead of here. That is a better place to reference older versions and is a better way of getting the attention of existing users.

JunruZhu: Hi, Lenny. We have changed the implementation in GLE-2819. Once it is merged, it won't be a problem. From my point of view, this warning is unnecessary now.

modules/API/pages/built-in-endpoints.adoc
```diff
@@ -739,148 +739,227 @@ $ curl -X GET 'http://localhost:9000/deleted_vertex_check?threadnum=10&verbose=0
...
+    "secret": <secret_value>, <1>
+    "graph": <graph_name>, <2>
+    "lifetime": <token_lifetime> <3>
+}
+----
+<1> User's secret to generate the token.
+Required if the request body does not supply `graph`.
+<2> Name of the graph that the token is valid for.
+Required if the request body does not supply `secret`.
+<3> Period of time for which the token is valid measured in seconds. The default value is about 2.6 million (about a month).
```
junhyungshinAtTigergraph, line 851: This part doesn't seem to be consistent with the example below. Please update accordingly.

modules/API/pages/built-in-endpoints.adoc
```diff
@@ -739,148 +739,227 @@ $ curl -X GET 'http://localhost:9000/deleted_vertex_check?threadnum=10&verbose=0
...
+----
+{
+    "secret": <secret_value>, <1>
+    "graph": <graph_name>, <2>
+}
+----
+<1> User's secret to generate the token.
+Required if the request body does not supply `graph`.
+<2> Name of the graph that the token is valid for.
+Required if the request body does not supply `secret`.
```
junhyungshinAtTigergraph, line 919: This part doesn't seem to be consistent with the example below. Please update accordingly.

junhyungshinAtTigergraph (commented):
I know the PR is already merged, but here are some necessary changes that we need to make.

PR #15, opened 2022-01-07
Ha update doc 1003

lennessyy:


lennessyy:
Merging now because the information is already outdated. Will add further edits if necessary.

lennessyy:
Hi @elliotmartinTG , I merged this PR because I know the content is more up-to-date than the old content even if there might be small inaccuracies. However, can you still take a look at the content and let me know if further edits are required? https://docs.tigergraph.com/tigergraph-server/3.2/ha/ha-cluster

PR #18, opened 2022-01-10
Doc 948 atomic request header

lennessyy:


modules/API/pages/built-in-endpoints.adoc
```diff
@@ -1297,9 +1297,10 @@ This endpoint upserts vertices and/or edges into a graph. To upsert means that i
 
 ==== Atomic upsert transaction
 
-By default, the `+POST /graph/{graph_name}+` endpoint is not atomic. If something goes wrong during the process of the request,  the request data can be partially consumed by the database.
+By default, the `+POST /graph/{graph_name}+` endpoint is not atomic. If something goes wrong during the process of the request, the request data can be partially consumed by the database.
 
-You can append a query string parameter `atomic_post` to the URL of the request and set its value to true to make the request an atomic transaction, which means that updates to the database contained in the request are all-or-nothing. Either all changes are successful, or none is successful.
+You can append a request header `gsql-atomic-level` to the request and set its value to `atomic` to make the request an atomic transaction.
```
aozeng, line 1302: Can you also add that the user have the option to specify that a request is non-atomic by setting header ```gsql-atomic-level``` to ```nonatomic```?

aozeng (changes requested):

modules/API/pages/built-in-endpoints.adoc
lennessyy: @aozeng I was under the impression that `nonatomic` is the default behavior. Do you just want to share this as an option? Or does this apply to some other atomic endpoints and can make them not atomic?

aozeng: Yes, I'd like to list the option to use `nonatomic` for `upsert`, even though it does not change the default behavior.

The user might want to explicitly use this `nonatomic` value in their REST request to stress that the choice of the  non atomic `upsert`  is intended and to inform the code viewer the existence of a header for atomic `upsert`.

lennessyy:
Hi @aozeng , I've added the `nonatomic` option

modules/API/pages/built-in-endpoints.adoc
```diff
@@ -1329,15 +1343,17 @@ And the content of `vertices.json` is:
...
 
-If you add the `atomic_post` parameter to the request URL and set its value to true, the request becomes atomic and if any part of the request body fails to be upserted, nothing will be upserted:
+If you add the `gsql-atomic-level` header to the request URL and set its value to `atomic`, the request becomes atomic and if any part of the request body fails to be upserted, nothing will be upserted:
 
-[source,bash]
+[source,console]
 ----
 # This is an atomic request
-curl --data-binary @vertices.json http://localhost:9000/graph/social
+$ curl -X POST -H 'gsql-atomic-level:atomic' --data-binary @vertices.json http://localhost:9000/graph/social?automic_post=true
```
aozeng, line 1356: Remove deprecated `automic_post` parameter.

lennessyy:
@aozeng thanks for catching that - I had removed it before but it came back after some merge conflicts. Fixed.

aozeng (approved):

PR #19, opened 2022-01-11
updated FAQ with error message

a-m-thomas:


lennessyy (approved):

modules/user-access/pages/ldap.adoc
```diff
@@ -29,10 +29,12 @@ To configure a TigerGraph system to use LDAP, there are two main configuration s
 . Configure the LDAP Connection.
 . Configure GSQL Proxy Groups and Users.
 
-In order to choose and specify your LDAP configuration settings, you must understand some basic LDAP concepts.  One reference for LDAP concepts is https://www.ldap.com/basic-ldap-concepts.
+In order to choose and specify your LDAP configuration settings, you should be familiar with some link:https://www.ldap.com/basic-ldap-concepts[basic LDAP concepts].
 
 === Prerequisites
 
+On a cluster, all nodes must have the LDAP certification file in the same place.
```
victor-gsl, line 36: Should this be 'certificate file' instead of 'certification file'? Neither term seems to be standard; it is our own creation to refer to the file in which the certificate resides.
P.S. I didn't see any reference to the name of the certificates. I assume that the files also have to have the same name, not just the same location. Perhaps we should specify that requirement as well.

victor-gsl (commented):
1 comment about the certification files (2 instances).

PR #27, opened 2022-01-24
Added notes about kafka loader | file format supported 

OmarAyo1:
Adding notes about what is the supported file format for kafka loader and the behaviour for wrong file format


OmarAyo1:
closing as so many many merge conflicts are reported. Will open the same for 3.2 branch

PR #29, opened 2022-01-25
changed input parameter data order

a-m-thomas:


lennessyy:
@a-m-thomas One thing you didn't mention was that the query parameters paper in the same order as they are in the query. I added it in a commit.

lennessyy (approved):

PR #38, opened 2022-02-04
edited GSQL client user doc

lennessyy:
Hi @duchoaile , I edited the remote GSQL client page,  and added instructions to retrieve the SSL certificate for cloud instances.

I also added a download link to GSQL client 3.4.0 so users don't have to install TigerGraph just to use the GSQL shell on the cloud. 

I will link to this page from the cloud docs, instead of duplicating the content. Please take a look and let me know if this looks good to you.


duchoaile (approved):
LGTM

PR #39, opened 2022-02-08
added release and patch process page

a-m-thomas:


modules/intro/pages/release-process.adoc
```diff
@@ -0,0 +1,99 @@
...
+
+*Notification:* Full releases will be announced on our documentation website.
+TigerGraph’s Customer Success team will also proactively share the release notes with existing customers.
+
+*Upgrades:* Customers using a legacy release can refer to the upgrade section of the online documentation to plan a server upgrade or contact TigerGraph Technical Support Engineering by opening a ticket at http://tigergraph.freshdesk.com[tigergraph.freshdesk.com] or emailing support@tigergraph.com.
+
+=== Patch Release
+
+A patch release is a release version that only includes fixes for the most recent one or two full releases.
+It will only be created when a set of critical bug fixes has accumulated that will affect a majority of customers before the next planned full release.
```
victor-gsl, line 64: Let's cut out "that will affect a majority of customers".  Including it sets up the negative idea that TigerGraph won't fix _my_ problem because I happen to be in the minority. It's enough to say the bugs are "critical", because our definition of _critical_ already embraces the impact to users in realistic settings.

victor-gsl (changes requested):
These are mostly editorial comments. Since this is a policy document, the policy and the wording should have been worked out before handing it a writing team.  Professional writers may have additional comments for improving how to express the intended ideas.

As part of the relevant leadership team, I did review and revise this document before it reached you. On my second review, I see additional things that I would change.

lennessyy (approved):

PR #42, opened 2022-02-21
clarify ACL import/export

lennessyy:


modules/import-export/pages/database-import-export.adoc
```diff
@@ -202,60 +226,57 @@ to enter a password when using this option.
 
 |`+-KU+` or `+--KEEP-USERS+` a|Keep the current users during the
 import operation. New users from the imported graph will still be added.
+Global roles from the current database are kept if you use this option.
```
QXG2, line 229: The privileges on local graphs are also cleared from global roles, since the local graphs are dropped.

QXG2 (approved):

PR #45, opened 2022-02-25
add note to reduce data skew

lennessyy:
- Added node for reducing data skew with an example


elliotmartinTG (approved):
Fantastic explanation. This looks great.

PR #50, opened 2022-03-15
Data streaming json doc 1054

lennessyy:
- Added note about JSON support

### DIFF
--- a/modules/data-loading/pages/data-streaming-connector.adoc
+++ b/modules/data-loading/pages/data-streaming-connector.adoc
@@ -56,12 +56,14 @@ The value for this configuration should be `com.google.cloud.hadoop.fs.gcs.Googl
 The streaming connector supports the following file types:
 
 * CSV files
+* JSON files.
+Each JSON object must be on a separate line.
 * directories
 * tar files
 * zip files
 
-For URIs that point to directories and compressed files, the connector will look for all CSV files inside the directory or the compressed file.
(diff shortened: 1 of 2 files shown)


modules/data-loading/pages/data-streaming-connector.adoc
```diff
@@ -56,12 +56,14 @@ The value for this configuration should be `com.google.cloud.hadoop.fs.gcs.Googl
...
 * CSV files
+* JSON files.
+Each JSON object must be on a separate line.
 * directories
 * tar files
 * zip files
 
-For URIs that point to directories and compressed files, the connector will look for all CSV files inside the directory or the compressed file.
-If you set `file.recursive`, the connector will look for CSV files recursively.
+For URIs that point to directories and compressed files, the connector looks for all CSV or JSON files inside the directory or the compressed file.
```
ryantigergraph, line 65: Not quite. It will try to load everything if file.regexp=".*", if file.regexp='.json', it will load json ONLY. You can also do regex expressions like ".(json|txt)" which will load json AND txt files.

ryantigergraph (changes requested):

modules/data-loading/pages/data-streaming-connector.adoc
lennessyy: Thanks for clarifying! I've clarified in the document

ryantigergraph (approved):

PR #55, opened 2022-03-25
clarify access requirements

a-m-thomas:


modules/installation/pages/install.adoc
```diff
@@ -115,12 +118,11 @@ image::image (77).png[Status of all components upon successful installation]
 
 === Non-interactive installation
 
-The following describes the procedure to install TigerGraph on Linux non-interactively.
+Non-interactive installation uses TigerGraph defaults for all prompts.
```
lennessyy, line 121: I think it's worth it to keep the sentence that describes the section at a high level, especially if the topic of the section is changing,

lennessyy (approved):

PR #59, opened 2022-04-01
added clarification that parameters are left blank

lennessyy:
Added note to clarify


JordanHorner (approved):
Looks good.  When we want to stop CrossRegionReplication, we don’t need to set IP/Port and they don’t need to change the topic prefix as It is already set.

PR #60, opened 2022-04-05
Pingfed doc 1075

lennessyy:
Added instructions for setting up SSO with PingFed.


jiemingzhou-tigergraph:
LGTM. Please consider adding PingFederate into supported IDP list.
image

PR #61, opened 2022-04-07
Upgrade requirements doc 1077

a-m-thomas:


lennessyy (approved):
Looks good. I made an edit to say "upgrade to another minor or major version" instead of just major version, since it's Major.Minor.Patch.

PR #70, opened 2022-04-26
Query status endpoints doc 1050

a-m-thomas:


modules/API/pages/built-in-endpoints.adoc
```diff
@@ -2932,6 +2932,8 @@ curl --fail -u <my_username>:<my_password> -X POST "http://localhost:14240/gsqls
 
 This endpoint reports statistics of running queries of a graph: the query's request ID, start time, expiration time, and the REST endpoint's URL.
 
+This endpoint only shows queries sent to the requested node, not all queries in the graph.
+
```
lennessyy, line 2936: "Node" should be in contrast to "cluster", not "graph". 

"If you are running a TigerGraph cluster, this endpoint only shows queries running on the node to which the request is sent, not all running queries on the cluster."

modules/API/pages/built-in-endpoints.adoc
```diff
@@ -3048,6 +3052,8 @@ Response::
 
 This endpoint allows you to check the status of a query run in xref:3.2@gsql-ref:querying:query-operations.adoc#_detached_mode_async_option[detached mode].
 
+This endpoint only allows you to check the status of a query sent to the requested node, not any query in the graph.
```
lennessyy, line 3055: Same edit re: graph vs cluster.

modules/API/pages/built-in-endpoints.adoc
```diff
@@ -3048,6 +3052,8 @@ Response::
 
 This endpoint allows you to check the status of a query run in xref:3.2@gsql-ref:querying:query-operations.adoc#_detached_mode_async_option[detached mode].
```
lennessyy, line 3053: Let's take these opportunities to remove references to the 3.2 version specifier. I added these by mistake when I first moved us to Antora and have been removing them when I see them ever since.

lennessyy (approved):

PR #76, opened 2022-05-04
Update built-in-endpoints.adoc

ElliotMtg2011:
Change `/ts3/api/datapoints:14240` to `:14240/ts3/api/datapoints` so that port comes before the API to match actual usage.


lennessyy (approved):

lennessyy:
Thanks for correcting!

PR #83, opened 2022-05-24
Avro support doc 1130

lennessyy:


modules/data-loading/partials/create-connector.adoc
```diff
@@ -0,0 +1,10 @@
+=== Create connector
+Once you have prepared the configuration file, run command `gadmin connector create` and specify the configuration file to create the connector:
+
+  gadmin connector create --c <config_file>
+
+The connectors start streaming from the data source immediately after creation if the configurations are valid.
+You can run `gadmin connector status` to verify the status of the connectors.
+If the configurations are valid, the connectors should be in `RUNNING` status.
+
+Data streamed from the source stays in TigerGraph's internal Kafka topics until they are ingested by a loading job.
```
ryantigergraph, line 10: The target kafka topic does not have to be tigergraph kafka, it can be any accessible kafka

modules/data-loading/partials/create-data-source.adoc
```diff
@@ -0,0 +1,23 @@
+=== Create data source
+Now that the connector has started loading data into TigerGraph's internal Kafka cluster, you can create a data source and point it to the Kafka cluster:
+
+. Create a data source configuration file.
+The broker's IP and hostname should be `localhost:30002`, which points to the port for TigerGraph's internal Kafka cluster.
```
ryantigergraph, line 5: Not necessarily localhost:30002 even for tigergraph kafka. It is only localhost when tigergraph is installed on single node cluster. It is usually the private IP address of the node and it can be retrieved using `gmyip`

modules/data-loading/pages/data-streaming-connector.adoc
```diff
@@ -400,12 +332,293 @@ GSQL > RUN LOADING JOB load_person
...
 
-==== Stream data source updates
+=== Stream data source updates
 
 To load updates to the data source after you've created the connectors, you need to delete the connector and recreate another connector.
 
 In the future, the streaming data connector will support automatically scanning for updates and stream data to TigerGraph.
 
+== Stream data from an external Kafka cluster
+You can create a data connector between an external Kafka cluster (source cluster) and TigerGraph’s internal Kafka server (destination cluster) to stream data from the source cluster.
```
ryantigergraph, line 342: Please rephrase this to something like: You can create a data connector stream data from an external Kafka cluster (source cluster) to TigerGraph’s internal Kafka server (destination cluster).
Also the destination does not have to be Tigergraph's internal Kafka server. I can be any kafka cluster

modules/data-loading/pages/data-streaming-connector.adoc
```diff
@@ -400,12 +332,293 @@ GSQL > RUN LOADING JOB load_person
...
+The connector streams the Avro data to TigerGraph's internal Kafka server, and transforms the data into JSON format.
```
ryantigergraph, line 343: The avro data transformation is done with our Avro converter which is a different component from our connectors. If users have csv in their external kafka, they can change the avro converter to bytearray converter and the csv format will remain and data will be streamed to the target kafka in csv.

modules/data-loading/pages/data-streaming-connector.adoc
```diff
@@ -400,12 +332,293 @@ GSQL > RUN LOADING JOB load_person
...
+You need to upload the following credential files to your TigerGraph cluster:
+
+* If the source cluster uses SASL, you need to upload the keytab of each Kerberos principal to every node of your TigerGraph cluster at the same absolute path.
+* If the source cluster uses SSL, you need to upload the trust store and key store to every node of your TigerGraph cluster.
+Each file must be at the same absolute path on all nodes.
+* If the source cluster uses SASL and SSL, you need to upload the keytab of each Kerberos principal, as well as the key store and truststore to every node of your TigerGraph cluster.
+Each file must be at the same absolute path on all nodes.
+
+The following configurations are required for admin, producer and consumer.
+To supply the configuration for the corresponding component, replace `<prefix>` with `admin`, `producer`, or `consumer`.
```
ryantigergraph, line 434: source.admin required

modules/data-loading/examples/config-avro
```diff
@@ -0,0 +1,57 @@
...
+replication.factor=1
+sync.topic.acls.enabled=false
+checkpoints.topic.replication.factor=1
+heartbeats.topic.replication.factor=1
+offset-syncs.topic.replication.factor=1
+offset.storage.replication.factor=1
+status.storage.replication.factor=1
+config.storage.replication.factor=1
+emit.heartbeats.interval.seconds=5
+secondary.scheduled.rebalance.max.delay.ms=35000
```
jinhui-tigergraph, line 17: @ryantigergraph pointed out the field should be prefixed with target cluster alias `world`. It should be
```
world.scheduled.rebalance.max.delay.ms=35000
```

modules/data-loading/pages/data-streaming-connector.adoc
```diff
@@ -400,12 +332,293 @@ GSQL > RUN LOADING JOB load_person
...
+The file in the example creates a connector named `avro-test-without-registry` with all configurations specified in the file.
+
+include::partial$create-connector.adoc[]
+
+include::partial$create-data-source.adoc[]
+
+include::partial$create-loading-job-kafka.adoc[]
+
+For example, suppose we have the following graph schema with only one vertex type
+ and the following JSON line data in Kafka:
```
jinhui-tigergraph, line 527: It would be better to make it explicit that the data is in tigergraph's Kafka instead of external one.

modules/data-loading/partials/create-connector.adoc
lennessyy: What would be the use case for users to stream to another Kafka cluster in this case? Is that something we want to cover?

ryantigergraph: Yes. Kafka loader can load data from external Kafka. 
A use case would be tigergraph kafka cluster does not have enough disk but have large memory and CPU, and they have another kafka cluster with huge disk to just store data. 
Double check, do we want to cover this case? cc @chengjie-qin

modules/data-loading/partials/create-data-source.adoc
```diff
@@ -0,0 +1,23 @@
...
+In the `kafka.config` field, set `group.id` to `tigergraph`:
```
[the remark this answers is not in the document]
lennessyy: For `kafka.config`, which section of the configuration settings in Kafka docs should I refer people to? Would it be Kafka Connect configs? https://kafka.apache.org/documentation/#configuration

lennessyy:
@ryantigergraph @jinhui-tigergraph Thanks for the feedback - I've addressed your comments. Please take a look

jinhuiztg (approved):

modules/data-loading/partials/create-connector.adoc
```diff
@@ -0,0 +1,13 @@
...
+[.wrap,console]
+----
+gadmin connector create --c <config_file>
+----
+
+The connectors start streaming from the data source immediately after creation if the configurations are valid.
+You can run `gadmin connector status` to verify the status of the connectors.
+If the configurations are valid, the connectors should be in `RUNNING` status.
+
+Data streamed from the source stays in the destination cluster Kafka topics until they are ingested by a loading job.
```
ryantigergraph, line 13: Do we plan to add more details on how our Pruner works? Will that be in a separate doc? We need to tell users about prune interval and some known issues for pruners.

alexthomasTG (approved):
Looks good, no changes

lennessyy:
@ryantigergraph I'll merge this first; if we have updates regarding message removal of non-TG internal kafka cluster, I'll make an update then

PR #88, opened 2022-06-02
initial draft of splitting commands and examples

alexthomasTG:


lennessyy (commented):
Nice work. This is a good start for us to start building out docs for more administrator-oriented tasks. 

The only thing I'm worried about is that `Management with gadmin` is too generic of a title. Is it possible for us to start dividing this into separate pages early on? I foresee us having to divide it over time

modules/reference/pages/configuration-parameters.adoc
```diff
@@ -839,9 +839,13 @@ System.LogRoot) of log directory for Nginx |`+nginx+`
 
 |Nginx.ConfigTemplate |The template to generate nginx config. Please use
 `+@filepath+` to parse template from file. Check the default template
-first at https://docs.tigergraph.com.(Warning: Don’t modify the reserved
+first here. (Warning: Don’t modify the reserved
```
lennessyy, line 842: This was auto generated from command-line help text. It's nice that you edited this, but the next time we generate it will still look like the original.

modules/system-management/pages/management-commands.adoc
```diff
@@ -0,0 +1,800 @@
+= Command glossary
+:description: List of commands for system management
+:page-aliases: gadmin:management-commands.adoc
```
lennessyy, line 3: I don't think this is the right alias? It should be the previous name of the page. I think this would be management-with-gadmin.adoc?

modules/system-management/pages/management-commands.adoc
```diff
@@ -0,0 +1,800 @@
...
+
+== Introduction
```
lennessyy, line 5: no need for the introduction heading if the page opens with the introduction

modules/system-management/pages/memory-management.adoc
```diff
@@ -79,7 +79,7 @@ There are three memory states:
 Healthy:: There is a healthy amount of free memory.
 All system operations run normally for best performance.
 Alert:: There is a shortage of free memory.
-TigerGraph's Graph Processing Engine (GPE) starts to slow down processing new requests to allow long-running queries to finish and release memory.
+TigerGraph's GPE starts to slow down processing new requests to allow long-running queries to finish and release memory.
```
lennessyy, line 82: We should spell out the full term when first mentioning it on a page. I think you can go back to the top of the page and spell out GPE.

lennessyy (approved):

lennessyy:
Remember you are merging to 3.5 only. Need to copy these changes to 3.6 in order to preserve them in future versions

PR #89, opened 2022-06-06
add note for downgrade

alexthomasTG:


ElliotMtg2011 (commented):
As an FYI, we are moving away from FreshDesk to ZenDesk. The FD article links will still be valid for the next couple of months, but will likely break after that. All of the articles will be moved to ZenDesk.

ElliotMtg2011 (approved):
LGTM

lennessyy (approved):

PR #104, opened 2022-07-14
add note about oracle installation

alexthomasTG:


lennessyy (approved):
Can we confirm that these steps also apply to 3.5? If so, we should propagate the changes to 3.5

a-m-thomas:
> Can we confirm that these steps also apply to 3.5? If so, we should propagate the changes to 3.5

Good point, I've asked Limin to review this.

LiminWangCS:
tested with 3.5.0 on OEL8, the install process also applies to it.

a-m-thomas:
cherry-picked to 3.5

PR #105, opened 2022-07-20
update client steps to reflect cloud use

alexthomasTG:


a-m-thomas:
I reworked the GSQL Client documentation into two sections, one for Server and one for Cloud. Even if we collapse the two different client versions into one, I think we should keep two separate links active.

elliotmartinTG:
We now have the client jar hosted at URLs of this form:
https://tigergraph-release-download.s3.us-west-1.amazonaws.com/gsql_client/3.6.1/gsql_client.jar

@arunramasami can you confirm this will be the long term location of client jars? If so, we should add these to this documentation update.

arunramasami:
I don't think this will be a long-term location. We (and Chengbiao) are
working on the location - this will also be updated on the download page.

@Chengbiao Jin ***@***.***>  - can you confirm?

Thanks,
Arun Ramasami
Principal Product Manager, Tigergraph

arunramasami:
It’s already on the download page. Please check and let me know if there is
any feedback.

Thanks,
Chengbiao

arunramasami:
Checkout here. https://dl.tigergraph.com/download.html

ElliotMtg2011 (changes requested):
Need to add info about client being available at dl.tigergraph.com

ElliotMtg2011 (approved):

modules/gsql-shell/pages/using-a-remote-gsql-client.adoc
```diff
@@ -73,6 +72,46 @@ alias gsql="java -jar <path_to_gsql_client>"
...
+=== Download the GSQL client
+
+link:https://dl.tigergraph.com/download.html[Download the GSQL client for a TigerGraph solution version directly from this page].
+
+=== Obtain SSL certificate
+Run the following command, replacing `<domain>` with the domain of your Cloud solution and replace `<path_to_certificate>` with the local path to place the certificate file.
+
+Find the domain of a Cloud solution by clicking on it in the solution view under "My Solutions."
+
+You need to install link:https://www.openssl.org/[`openssl`] on your client machine to run this command.
```
lennessyy, line 88: Should add this to prerequisites

lennessyy (approved):

PR #107, opened 2022-08-02
order ports logically and add note about internal and external ports

alexthomasTG:


modules/reference/pages/ports.adoc
```diff
@@ -1,121 +1,122 @@
 = List of Ports
 :description: List of all ports on TigerGraph.
 
+This page lists all ports that TigerGraph uses.
+
+TigerGraph requires three external ports to be exposed: 22, 9000, and 14240. All other ports are used for internal communication within a cluster.
```
lennessyy, line 6: Should note that the other ports need to be open within the cluster

lennessyy (approved):

PR #113, opened 2022-08-10
Max timeout doc 1264

alexthomasTG:


modules/API/pages/index.adoc
```diff
@@ -281,20 +280,25 @@ The maximum size for a request body, including the payload file, is set by the s
...
+To customize this timeout limit for a particular query instance, set the `GSQL-TIMEOUT` parameter in the request header.
+If you are using `curl` to submit your RESTPP request, the syntax would be the following:
 
 [source,bash]
 ----
 curl -X <GET/POST> -H "GSQL-TIMEOUT: <timeout value in ms>" '<request_URL>'
 ----
 
+The longest supported timeout is 2147483647 or stem:[2^31-1] seconds (68 years).
+If a larger timeout value is entered, the value will fall back to the default of 16 seconds.
```
anthonywilson-TG, line 300: To be more specific, if a value larger than 2147483647 is entered, it will fall back to the value currently configured for the RESTPP.Factory.DefaultQueryTimeoutSec parameter. The default after installation is 16 seconds, but several customers change this. In the case this was discovered, it was for Wayfair and they had it set to 3600 (1 hour)

anthonywilson-TG (commented):
To be more specific, if a value larger than 2147483647 is entered, it will fall back to the value currently configured for the RESTPP.Factory.DefaultQueryTimeoutSec parameter. The default after installation is 16 seconds, but several customers change this. In the case this was discovered, it was for Wayfair and they had it set to 3600 (1 hour)

lennessyy (approved):

PR #125, opened 2022-09-06
Password policy doc 1335

lennessyy:


modules/security/pages/password-policy.adoc
```diff
@@ -0,0 +1,30 @@
...
+To enable the password policy, run the following commands:
+
+[.wrap,console]
+----
+$ gadmin config set Security.UserPasswordPolicy.Enable true <1>
+$ gadmin config set Security.ACLPasswordPolicy.Enable true  <2>
+$ gadmin config apply
+$ gadmin restart gsql
+----
+<1> Setting `Security.UserPasswordPolicy.Enable` to true applies the password policy to passwords for TigerGraph database users.
```
QXG2, line 29: also add them to https://docs.tigergraph.com/tigergraph-server/current/reference/configuration-parameters

And mention that the policy won't disable existing user with simple password.

lennessyy: yes, they'll be added as I run a script to produce all the new config parameters for every version

QXG2: Still need to mention that
* the policy won't disable existing user with simple password.
* mandatory password policy than always exist: password can't be empty string ("") or "null".

lennessyy: Added to the draft

alexthomasTG (approved):

PR #132, opened 2022-09-20
Expire token removal doc 1344

lennessyy:
The option to provide users and secrets through query string parameters is omitted because of security concerns. I feel like we might end up removing the ability to do this at all in the future so it's best not to document it now in case we have to deprecate it.

### DIFF
--- a/modules/API/pages/built-in-endpoints.adoc
+++ b/modules/API/pages/built-in-endpoints.adoc
@@ -1113,6 +1113,9 @@ If not provided a request body, a request deletes all expired tokens.
 The secrets and users provided in the request body do not have to correlate with each other.
 A request deletes all tokens associated with any of the users and secrets contained in the payload.
 
+The request is atomic:
+if any of the provided users or secrets are invalid, or if a user doesn't have the privilege to remove tokens that belong to one of the users or secrets provided, no tokens will be removed by the request.
+
 ==== Parameters
 None.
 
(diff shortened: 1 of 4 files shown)


modules/API/pages/built-in-endpoints.adoc
```diff
@@ -1070,6 +1070,55 @@ Response::
...
+{
+    "user": ["string"], <1>
+    "secret": ["string"] <2>
+}
+----
+<1> Users whose expired tokens to remove. Optional.
+<2> If a token is created with one of the secrets in the list, they are removed by the request. Optional.
+
+The secrets and users provided in the request body do not have to correlate with each other.
+A request deletes all tokens associated with any of the users and secrets contained in the payload.
```
qianzhang-tg, line 1114: It might worth to mention that this endpoint behavior is atomic and illustrate what happens if any of payload parameter is non-exist, invalid or the user does not have sufficient privilege.
Essentially, if any user or secret in the payload is invalid/non-exist, or the current user does not have enough privilege on some user/secret, then the request will fail and none of the expired tokens will be deleted.

lennessyy: Added the clarification

qianzhang-tg:
Looks good to me.

alexthomasTG (approved):

PR #150, opened 2022-10-25
Doc 1429 multiedge

lennessyy:


modules/API/pages/built-in-endpoints.adoc
```diff
@@ -2606,21 +2178,78 @@ Response::
...
+----
+curl -X GET 'http://localhost:9000/graph/multiedge/edges/Person/4/transfer/Person/30/2431'
+----
+--
+Composite discriminator::
++
+--
+[.wrap,console]
+----
+curl -X GET "http://localhost:9000/graph/multiedge/edges/Person/9/transfer2/Account/9/77,1117,2010-04-09%2003:57:01" <1>
```
dimu637, line 2219: if the discriminator has multiple attributes, use ```,``` as a separator.
But, if the attributes already have a ```,``` in a string discriminator, use ```\``` as an escape character.
eg
```
st\,\,r\,1,101,2012-02-12%2011:32:00
```
means 
```
attr1: st,,r,1
attr2: 101
attr3: 2012-02-12%2011:32:00
```
Also,  ```\``` can escape  ```\```
eg:
```
"\\st\\\,r\,\\\,,101,2012-02-12 11:32:00"
```
means
```
attr1: \st\,r,\,
attr2: 101
attr3: 2012-02-12%2011:32:00

```

modules/API/pages/built-in-endpoints.adoc
```diff
@@ -2664,18 +2293,68 @@ Response::
...
+----
+curl -X DELETE 'http://localhost:9000/graph/multiedge/edges/Person/4/transfer/Person/30/2431'
+----
+--
+Composite discriminator::
++
+--
+[.wrap,console]
+----
+curl -X DELETE "http://localhost:9000/graph/multiedge/edges/Person/9/transfer2/Account/9/77,1117,2010-04-09%2003:57:01" <1>
```
dimu637, line 2331: The DateTime discriminator attribute can be a string format(string eg 2010-04-09%2003:57:01) or an epoach time(unit, eg 1666892012)

modules/API/pages/built-in-endpoints.adoc
dimu637, line 2219: some expected error message will be 

wrong type:
```
"Discriminator "str2" cannot be converted into UINT.
```
```
Discriminator "str1" cannot be converted into DATETIME.
```
more/less discriminator provided
```
The number of the input discriminators is 2. But the number of discriminators in schema is 3.
```
```
The input discriminator has more fields than the specified in the schema. 3 discriminators are parsed at position 28.
```

lennessyy: I added the explanation but not the example, because it is kind of a low-likelihood with too intrusive of an explanation.

lennessyy: Usually, it is useful to document an error message if it isn't self-explanatory. For example, something like "Segmentation fault" is not explanatory and it can be helpful to provide an explanation. 

I don't think it is necessary to document error messages here, as the error messages themselves seem to be self-explanatory.

alexthomasTG (approved):

PR #162, opened 2022-11-11
add paragraph on queue

lennessyy:


modules/system-management/pages/workload-management.adoc
```diff
@@ -29,6 +29,17 @@ For example, to specify that there can only be 50 concurrent queries at a time,
 $ gadmin confit set RESTPP.WorkLoadManager.MaxConcurrentQueries 50
 ----
 
+If the maximum number of concurrent queries is reached, newly submitted queries are placed in a queue, and begin to run as the currently running queries finish.
```
venu-joshi, line 32: prefix queue with "Delay" as in "Delay Queue"

venu-joshi (approved):

PR #169, opened 2022-12-01
clarify memory alert and tweak wording

alexthomasTG:


modules/system-management/pages/memory-management.adoc
```diff
@@ -123,7 +122,8 @@ There are three memory states:
 Healthy:: There is a healthy amount of free memory.
 All system operations run normally for best performance.
 Alert:: There is a shortage of free memory.
-TigerGraph's GPE starts to slow down processing new requests to allow long-running queries to finish and release memory.
+TigerGraph's GPE system enables mechanisms to alleviate the memory pressure by moving some data onto disks and restricting the maximum allowed threads for rebuilding.
```
ghost: the code would try to use maximum allowed threads allowed for rebuilding instead of limiting it.

PR #170, opened 2022-12-01
add zmq keepalive

alexthomasTG:


ElliotMtg2011 (changes requested):
We want customer to leave this enabled. Setting to 0 will disable this, so we shouldn't make this recommendation. 

I think we'd be better served by leaving this here as the default value, and adding a separate page for documenting GPE environmental variables.

alexthomasTG:
I switched it to the default value of 1 and opened a new ticket for creating a GPE environment variables page.

PR #173, opened 2022-12-06
Update default value for ack in RESTPP POST

ElliotMtg2011:
Details based on discussion in EngineTeam Zulip stream: https://zulip.graphtiger.com/#narrow/stream/322-EngineTeam/topic/.28no.20topic.29/near/2471352 Per Songting, none should rarely be used and all is the default value in the code.

### DIFF
--- a/modules/API/pages/upsert-rest.adoc
+++ b/modules/API/pages/upsert-rest.adoc
@@ -25,7 +25,7 @@ acknowledged the POST request
 * `"none"`: request will return immediately after
 RESTPP processed the POST request.
 
-Default value is `"all"`. `"none"` is not recommended except in very specific cases as it could easily lead to overwhelming the system. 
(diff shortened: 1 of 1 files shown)


alexthomasTG (commented):
If we say "in very specific cases," can we add detail about what types of cases those might be? Otherwise, maybe we could just say " "none" is not generally recommended as it could easily lead to overwhelming the system. Contact TigerGraph Support if you believe your situation requires setting this to "none". "

ElliotMtg2011:
> If we say "in very specific cases," can we add detail about what types of cases those might be? Otherwise, maybe we could just say " "none" is not generally recommended as it could easily lead to overwhelming the system. Contact TigerGraph Support if you believe your situation requires setting this to "none". "

Thanks, good call. I've made the appropriate changes.

alexthomasTG (approved):

PR #185, opened 2023-01-12
update page with new information

alexthomasTG:


modules/crr/pages/set-up-crr.adoc
```diff
@@ -32,17 +32,65 @@ Make sure TigerGraph is *not* installed with a local loopback IP such as 127.0.0
...
+Use GBAR to xref:backup-and-restore:index.adoc[create a backup] of the primary cluster.
+See xref:backup-and-restore:index.adoc[Backup and Restore] for how to create a backup.
 
 If you are setting up both the primary cluster and the DR cluster from scratch, you can skip Steps 1, 2, and 4 and only perform Step 3.
 
 === Restore on the DR cluster
 
 Copy the backup files from every node to every node on the new cluster.  xref:backup-and-restore:index.adoc#_restore_from_a_backup_archive[Restore the backup] of the primary cluster on the DR cluster. See xref:backup-and-restore:index.adoc[Backup and Restore] on how to restore a backup.
 
+=== Delete metadata records
```
qianzhang-tg, line 46: Arun suggested to add a dialogue box to provide subversion details on which subversions needs to do this `Delete metadata records`. 
Currently, there is a fix that will be applied to 3.5.2+, 3.6.3+, 3.7.1+, 3.9.0+, 4.0+.
For specific subversions earlier than them, actions in this section are needed. Otherwise, user should not execute this section.

modules/crr/pages/set-up-crr.adoc
```diff
@@ -32,17 +32,65 @@ Make sure TigerGraph is *not* installed with a local loopback IP such as 127.0.0
...
+  ],
+  "version": 1
+}
+----
+
+`<crr metadata topic name>` is the topic name that CRR consumes.
+It contains a prefix determined by the gadmin config entry `System.CrossRegionReplication.TopicPrefix`.
+By default, this prefix is an empty string.
+
+Example: If System.CrossRegionReplication.TopicPrefix is set to Primary, then metadata topic used by CRR on DR cluster will be named Primary.Metadata. Therefore, users need to pay attention when filling the topic name in the JSON config.
```
qianzhang-tg: Wrap "System.CrossRegionReplication.TopicPrefix" into code font like `System.CrossRegionReplication.TopicPrefix`.
Same for `Primary`. 

I also would like to rephrase the example like below, maybe using a table?
> Example: 
> If `System.CrossRegionReplication.TopicPrefix` is empty string on the cluster, then `<crr metadata topic name>` is `Metadata`. 
> If `System.CrossRegionReplication.TopicPrefix` is `Primary` on the cluster, then `<crr metadata topic name>` is `Primary.Metadata`.

modules/crr/pages/set-up-crr.adoc
```diff
@@ -32,17 +32,65 @@ Make sure TigerGraph is *not* installed with a local loopback IP such as 127.0.0
...
+      "offset": -1
+    }
+  ],
+  "version": 1
+}
+----
+
+`<crr metadata topic name>` is the topic name that CRR consumes.
+It contains a prefix determined by the gadmin config entry `System.CrossRegionReplication.TopicPrefix`.
+By default, this prefix is an empty string.
```
qianzhang-tg: I would like to rephrase like the following:
> `<crr metadata topic name>` is the name of Kafka topic from which CRR consumes metadata messages. It is generated by concatenating  a topic prefix with `Metadata`. The topic prefix is determined by gadmin config entry `System.CrossRegionReplication.TopicPrefix`. Please refer to example below.

modules/crr/pages/set-up-crr.adoc
```diff
@@ -32,17 +32,65 @@ Make sure TigerGraph is *not* installed with a local loopback IP such as 127.0.0
...
 === Restore on the DR cluster
 
 Copy the backup files from every node to every node on the new cluster.  xref:backup-and-restore:index.adoc#_restore_from_a_backup_archive[Restore the backup] of the primary cluster on the DR cluster. See xref:backup-and-restore:index.adoc[Backup and Restore] on how to restore a backup.
 
+=== Delete metadata records
+
+After the restoration, but before enabling CRR on the DR cluster, delete the oudated Kafka metadata records.
+
+Kafka provides a script called `kafka-delete-records.sh`.
+It takes a JSON configuration file to specify which records/messages on which partition, and until which offset, you want to delete.
```
qianzhang-tg, line 55: Also at here, mention that "If offset is set to `-1`, the script will delete all records in the given partition."

modules/crr/pages/set-up-crr.adoc
alexthomasTG: Can you confirm that "actions in this section are needed" refers to the entire 2.3 "Delete metadata records" section?

qianzhang-tg: @alexthomasTG  Yes.

PR #187, opened 2023-01-14
add new endpoint

alexthomasTG:


modules/API/pages/built-in-endpoints.adoc
```diff
@@ -3184,7 +3184,63 @@ Response::
 ==== Parameters
 No parameters.
 
+=== Show all delayed queries
 
+`GET /showdelayedlistall`
+
+When a query request is sent while another query is still running, it is sent to a "delayed" queue.
```
venu-joshi: Not quite. 
>When queries beyond the RESTPP.WorkLoadManager.MaxConcurrentQueries are submitted to the TG system, these are gate into the delay or overflow queue. This overflow queue length is limited by RESTPP.WorkLoadManager.MaxDelayQueueSize parameter. Use this endpoint to get a list of all queries waiting in the delay/overflow queue.

modules/API/pages/built-in-endpoints.adoc
```diff
@@ -3184,7 +3184,63 @@ Response::
...
+    },
+    {
+      "requestidPartial": "1670970545942",
+      "startTime": "2022-12-13 22:29:05.942",
+      "url": "/query/ldbc_snb/LTQ1?gap=60&qtf=vwlm_a_004",
+      "elapsedTime": 2023
+    },
+  ],
+  "code": "REST-0000"
+}
```
venu-joshi, line 3238: Do we need consistency in the example responses.   Can we use the same graph/query examples as in /showprocesslistall documentation? https://docs.tigergraph.com/tigergraph-server/current/api/built-in-endpoints

venu-joshi (approved):
Please include the comments.

modules/API/pages/built-in-endpoints.adoc
venu-joshi: Can we change the query names to      
      "url": "/query/Work_Net/rngExample",
         "url": "/query/Entity_Resolution/rngExample",
     
     To be consistent across /showprocesslistall and /showdelayedlistall ?

PR #188, opened 2023-01-17
Data loading overhaul doc 1564

alexthomasTG:


modules/data-loading/pages/data-streaming-connector/kafka.adoc
```diff
@@ -85,6 +85,10 @@ This controls the format of the values in messages written to or read from Kafka
 URL of the schema registry service.
 Required if  `value.converter` is `com.tigergraph.kafka.connect.converters.TigerGraphAvroConverter`.
 For details, see https://docs.confluent.io/platform/current/schema-registry/connect.html#example-converter-properties[Confluent documentation].|`\http://localhost:8081`
+
+| `sync.topic.configs.enabled`
+|Whether to replicate topic configurations from the source cluster (default: true).
+Usually set to `false` as the config of the source Kafka topic is not necessary to sync data.
```
victorleeTG: I'm confused why the default is true, but we say "Usually set to `false`.  Who is the actor that usually sets this to false?

modules/data-loading/pages/data-streaming-connector/index.adoc
```diff
@@ -108,7 +108,151 @@ Use the `-s` or `--soft` flag to delete the connector only and leave the Kafka m
...
+=== Run the loading job
+
+Use the command `RUN LOADING JOB` to run the loading job.
+
+[source, gsql]
+----
+RUN LOADING JOB stream_csv
+----
+
+By default, new files or new lines in files in an external data source are not automatically loaded into TigerGraph once the loading job stops.
```
victorleeTG: It can be unclear what is considered "new". Does it mean anything added after the loading job starts?

modules/data-loading/pages/data-streaming-connector/index.adoc
```diff
@@ -108,7 +108,151 @@ Use the `-s` or `--soft` flag to delete the connector only and leave the Kafka m
...
+The data streaming connector also supports continuous loading with an `EOF` flag, short for "terminate loading at the end of the file."
```
victorleeTG: I find this use use of "short for" surprising:
* EOF as an acronym: yes, EOF is short for "end of file"
* EOF as the name of a flag: Every flag for every command is "short for" something, but we don't usually explain it that way. We simply explain the behavior, such as "If EOF is true, the continuous loading will stop when the loader encounters an end-of-file (EOF) character in the data."

modules/data-loading/pages/data-streaming-connector/index.adoc
```diff
@@ -108,7 +108,151 @@ Use the `-s` or `--soft` flag to delete the connector only and leave the Kafka m
...
+== Create the loading job
 
+The loading job creation happens entirely in GSQL.
+
+=== Specify a graph and config file
+
+[source.wrap, gsql]
+----
+USE GRAPH test_graph
+CREATE DATA_SOURCE STREAM s1 = "ds_config.json" for graph test_graph
```
victorleeTG:
```suggestion
CREATE DATA_SOURCE STREAM s1 = "ds_config.json" FOR GRAPH test_graph
```

modules/data-loading/pages/data-streaming-connector/index.adoc
```diff
@@ -108,7 +108,151 @@ Use the `-s` or `--soft` flag to delete the connector only and leave the Kafka m
 $ gadmin connector delete <connector_name> -s
 ----
 
+== Create the loading job
 
+The loading job creation happens entirely in GSQL.
```
victorleeTG:
```suggestion
Loading jobs for the Data Streaming Connector are created and run in GSQL.
```

modules/data-loading/pages/data-streaming-connector/index.adoc
```diff
@@ -108,7 +108,151 @@ Use the `-s` or `--soft` flag to delete the connector only and leave the Kafka m
...
+[source.wrap, gsql]
+----
+USE GRAPH test_graph
+CREATE DATA_SOURCE STREAM s1 = "ds_config.json" for graph test_graph
+----
+
+In this example, we create a data source of type `STREAM` and name it `s1`.
+We set our data source as the `ds_config.json` file.
+
+.JSON config file examples
```
victorleeTG, line 79: I found the presentation of this example confusing.  You show bits and pieces of what is needed, and then the entire job in the next subsection "Sample loading job syntax".  
I think it would be better to either
1. **Talk** about the high-level workflow (create a data_source stream, create a loading job, run the job), then drill down into details of each, or
2. **Show** and entire example workflow, then drill down to talk about the details.

modules/data-loading/pages/data-streaming-connector/index.adoc
```diff
@@ -108,7 +108,151 @@ Use the `-s` or `--soft` flag to delete the connector only and leave the Kafka m
...
+=== Specify a graph and config file
+
+[source.wrap, gsql]
+----
+USE GRAPH test_graph
+CREATE DATA_SOURCE STREAM s1 = "ds_config.json" for graph test_graph
+----
+
+In this example, we create a data source of type `STREAM` and name it `s1`.
+We set our data source as the `ds_config.json` file.
```
victorleeTG: ds_config.json is not the data source; it is the steam's config file.
That confusion is the fault of the command syntax.  The syntax should not be 
`STREAM s1 = <config_file>`
An example of a semantically correct syntax is
`STREAM s1 WITH CONFIG <config_file>`

victorleeTG (changes requested):

PR #191, opened 2023-01-19
add new env variable

alexthomasTG:


modules/reference/pages/configuration-parameters.adoc
```diff
@@ -892,7 +892,7 @@ the key from file) |`nan`
 |===
 |Name |Description |Example
 |RESTPP.BasicConfig.Env |The runtime environment variables, separated by
-`;' |`LD_PRELOAD=$LD_PRELOAD; LD_LIBRARY_PATH=$LD_LIBRARY_PATH;`
+`;' |`LD_PRELOAD=$LD_PRELOAD; LD_LIBRARY_PATH=$LD_LIBRARY_PATH;REPORT_FIRST_N_LINES = $REPORT_FIRST_N_LINES`
```
minzhangtg: The value of A is a non-negative integer.
A little formatting problem: We'd better keep the format consistent, with space after the semicolon and no space on either side of the equal sign.

PR #199, opened 2023-02-09
Update index.adoc

yuchenZhangTG:


modules/data-loading/pages/data-streaming-connector/index.adoc
```diff
@@ -181,48 +192,109 @@ LOAD file_Comment
...
+DEFINE FILENAME bq_inline_json = """$s1:myfile.json""";
+----
+where the content in `myfile.json` following the same format as the json content.
+
+--
+
+====
+
+=== Creating the loading job
+The instruction for creating loading job is the same as file loading [link]. Note:
```
yuchenZhangTG, line 281: Could you replace the link with the one to the create loading job page.

modules/data-loading/pages/data-streaming-connector/index.adoc
```diff
@@ -308,4 +380,4 @@ If the loading job finishes much sooner before the interval is reached, the mess
 * If a topic is deleted and recreated while a loading job on the topic is running, the data in the topic may get removed.
```
yuchenZhangTG, line 380: I think the known issues are fixed. We can remove all those known issue. We may add a section about "how stream loading works" and "management of connectors"
See the last 3 sections in https://graphsql.atlassian.net/browse/DOC-1608

PR #200, opened 2023-02-15
DOC-1658 Installer docs incorrectly state that sudo is not required f…

ElliotMtg2011:
…or single node install

### DIFF
--- a/modules/installation/pages/bare-metal-install.adoc
+++ b/modules/installation/pages/bare-metal-install.adoc
@@ -177,7 +177,8 @@ Navigate to the `tigergraph-<version>-offline`` folder and run the `install.sh` 
 ----
 $ cd tigergraph-<version>-offline
 
-#sudo not requried for standalone installation
+#sudo required for single node install
+#not required for clutser install, but sudo user must be supplied in AdvancedConfig
 $ sudo ./install.sh
 ----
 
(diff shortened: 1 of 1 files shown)


modules/installation/pages/bare-metal-install.adoc
```diff
@@ -177,7 +177,8 @@ Navigate to the `tigergraph-<version>-offline`` folder and run the `install.sh`
 ----
 $ cd tigergraph-<version>-offline
 
-#sudo not requried for standalone installation
+#sudo required for single node install
+#not required for clutser install, but sudo user must be supplied in AdvancedConfig
```
victorleeTG: #not required for clutser install, but sudo user must be supplied in AdvancedConfig
@ElliotMtg2011 , I didn't know what you meant by "AdvancedConfig". I searched our documentation for that.  It does not appear anywhere in 3.8 or 3.7 documentation, but it was mentioned in any older, more detailed 3.6 installation guide https://docs.tigergraph.com/tigergraph-server/3.6/installation/install#_step_2_configure_installation_settings

victorleeTG: I don't know why we removed this more advanced installation information from our documentation.

victorleeTG (changes requested):

modules/installation/pages/bare-metal-install.adoc
ElliotMtg2011: This refers to the install_conf.json JSON field. I updated the comment to make it clearer and fixed the typo.  

The same information is available in 3.7+:
https://docs.tigergraph.com/tigergraph-server/current/installation/bare-metal-install#_non_interactive_installation

victorleeTG (approved):

PR #203, opened 2023-02-23
update operator installation notes

alexthomasTG:


modules/kubernetes/pages/k8s-operator/cluster-operations.adoc
```diff
@@ -19,26 +19,54 @@ $ kubectl tg create --cluster-name <cluster_name>  --storage-class <storage_clas
 Each cluster is provisioned with https://kubernetes.io/docs/concepts/storage/persistent-volumes/[persistent volumes] for each node by default.
 When you create a cluster with persistent volumes, you must specify a https://kubernetes.io/docs/concepts/storage/storage-classes/[storage class].
 
+Run `kubectl get storageclass` to show the names of the available storage classes.
```
jerryyangtg, line 22: I think we should improve the part of creating a cluster, currently, it doesn't mention the private ssh key, and we can also remove the --private-key-secret part at LOC 61. @alexthomasTG you can check out the latest internal document, I guess you miss the latest update.

PR #204, opened 2023-02-27
DOC-1642 respectful terminology, DOC-1678 HA GSQL wording

victorleeTG:
Significant changes on the HA for GSQL Server page:
1. Rewrote the intro which was confusing and dated. I _think_ my new version is factually correct.
2. Moved a section from the bottom to the middle, to be a part of "GSQL Server HA Configuration".
3. Dropped a subsection that said (I think, because it was poorly worded) "We're going to add support for (local file something-or-other)", but we don't have that feature yet.

The Terminology change was changing "dummy" to "default".

### DIFF
--- a/modules/ha/pages/ha-for-gsql-server.adoc
+++ b/modules/ha/pages/ha-for-gsql-server.adoc
@@ -305,7 +305,7 @@ CREATE LOADING JOB load_kafka FOR GRAPH poc_graph {
 
 === GSQL Client connection setup
 
-The GSQL client can connect to the GSQL server in one of several ways according to following priority order:
+The GSQL client can connect to the GSQL server in one of several ways according to the following priority order:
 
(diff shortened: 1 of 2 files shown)


modules/ha/pages/ha-for-gsql-server.adoc
```diff
@@ -299,18 +305,18 @@ CREATE LOADING JOB load_kafka FOR GRAPH poc_graph {
 
 === GSQL Client connection setup
 
-The GSQL client can connect to GSQL server in the different ways with the following priority order:
+The GSQL client can connect to the GSQL server in one of several ways according to following priority order:
```
alexthomasTG:
```suggestion
The GSQL client can connect to the GSQL server in one of several ways according to the following priority order:
```

PR #227, opened 2023-03-23
Fix admin table structural error

victorleeTG:
Restore a cell that was accidentally deleted from the Admin table, which misaligned the entire table.


victorleeTG (commented):
looks good

PR #233, opened 2023-03-28
DOC-1713-security-change-password for docker and bare metal

victorleeTG:


modules/getting-started/pages/docker.adoc
```diff
@@ -134,6 +134,19 @@ $ docker exec -it $(docker ps | grep tigergraph | awk '{print $1}') /bin/sh
 $ gadmin start all
 ----
 
+. Attend to security.
```
venu-joshi, line 137: What does "attend to security" mean?   Just thinking if this should be a separate item before the "operations command cheat sheet" item.

== Securing your Docker container

WARNING: ....

modules/getting-started/pages/docker.adoc
```diff
@@ -134,6 +134,19 @@ $ docker exec -it $(docker ps | grep tigergraph | awk '{print $1}') /bin/sh
...
++
+WARNING: Your TigerGraph image is preconfigured with a superuser having the default username and password `tigergraph`. If you do not change this, anyone with access to your database will be able to read and modify it.
++
+To xref:tigergraph-server:user-access:user-management.adoc#_change_a_users_password[change the password] of the user called `tigergraph`:
```
venu-joshi, line 141: There are two passwords to change,
Linux user "tigergraph" with "tigergraph" password.  We can point them to the linux utility 'passwd' or just mention to use the Operating System tools to change the default password.
TigerGraph Application user "tigergraph" with "tigergraph" password.   Pointing them to "TigerGraph database user management" link is nice.

modules/getting-started/pages/docker.adoc
```diff
@@ -134,6 +134,19 @@ $ docker exec -it $(docker ps | grep tigergraph | awk '{print $1}') /bin/sh
...
++
+[source,console]
+----
+$ gadmin ALTER PASSWORD tigergraph
+----
++
+For additional security settings, see xref:tigergraph-server:security:index.adoc[].
```
venu-joshi, line 148: Also add note 
Please follow best practices for securing and hardening the docker container especially when installing in a shared environment.

venu-joshi (approved):
Minor comments/suggestions provided.

modules/getting-started/pages/docker.adoc
victorleeTG: At a high level, security should come before playing with your new DB, but I assume that the user may not be able config security until after they run `gadmin start all`.  Maybe that is incorrect.

modules/getting-started/pages/docker.adoc
victorleeTG: I don't disagree with you. I'm trying to find a phrasing that is directive for users who SHOULD think about security, which also understanding that a high percentage of users who "try" TigerGraph with Docker are working as individuals and don't care about protecting this particular DB instance.

modules/getting-started/pages/docker.adoc
victorleeTG: Ok. The need to change both users (Linux and TigerGraph DB) was not clear to me, so I didn't write that.

modules/getting-started/pages/docker.adoc
venu-joshi: The page is meant for people who install TG Software. If this is a user is performing an install that is going to be shared with additional users (over network etc.,), then they must follow "best practices". 
If this is a standalone user installing it over a private network/desktop with no access to external users, then this is not so important at the moment of 'let-me-try' but it becomes important when the try-environment starts getting shared with additional users.

PR #261, opened 2023-05-10
Fix port typo

rrossmiller:
It looks like there is a typo in the remote client documentation. It tells you to use port `12420`, but it seems, in practice, that port is still `14240` -- the same it has been in previous versions.


victorleeTG:
Thank you for your contributions!
Besides fixing 3.9, we will fix past versions back to 3.6 LTS.

victorleeTG (approved):
Looks good, thanks

PR #265, opened 2023-06-08
DOC-1722-change-informant-endpoints

victorleeTG:
By using 14240/informant/metrics instead of 9167/metrics, the document is simplied.
It uses the same port as many other endpoints, so the user will not be surprised.
Also, I think the special comments about the port are not needed.

Current PR is for 3.9.2.
It will then be backported to earlier versions.


chengjie-qin (approved):
LGTM. Thx for the change!

haoweig (approved):

PR #325, opened 2023-09-21
DOC-1862-fixed-5ClusterMaxTo3Cluster

JosephNewmanTG:
Associated task: https://graphsql.atlassian.net/browse/DOC-1862

Correction in docs for 5 max clusters to 3 max clusters. Plus style changes for clarity. To be merged in 3.9LTS, 3.8, 3.7, 3.6LTS.

Screen Shot 2023-09-22 at 10 57 42 AM

### DIFF
--- a/modules/cluster-and-ha-management/pages/ha-for-application-server.adoc
+++ b/modules/cluster-and-ha-management/pages/ha-for-application-server.adoc
@@ -5,7 +5,7 @@
 TigerGraph supports native HA functionality for its application server, which serves the APIs for TigerGraph's GUI - GraphStudio and Admin Portal.
 The application server follows the active-active architecture, and runs on the *first three nodes* in a cluster by default.
 
-NOTE: _**Three ** is the maximum number of application servers per cluster.
+NOTE: _Three is the maximum number of application servers per cluster.
 If one server falls offline, you can use the other servers without any loss of functionality._
 
 When you deploy TigerGraph in a cluster with multiple replicas, it is ideal to set up load balancing to distribute network traffic evenly across the different servers.
--- a/modules/cluster-and-ha-management/pages/ha-for-application-server.adoc
+++ b/modules/cluster-and-ha-management/pages/ha-for-application-server.adoc
@@ -3,7 +3,7 @@
 :description: Overview of high availability support for the application server.
 
 TigerGraph supports native HA functionality for its application server, which serves the APIs for TigerGraph's GUI - GraphStudio and Admin Portal.
-The application server follows the active-active architecture, and runs on the *first five nodes* in a cluster by default.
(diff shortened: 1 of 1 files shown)


modules/cluster-and-ha-management/pages/ha-for-application-server.adoc
```diff
@@ -3,11 +3,13 @@
 :description: Overview of high availability support for the application server.
 
 TigerGraph supports native HA functionality for its application server, which serves the APIs for TigerGraph's GUI - GraphStudio and Admin Portal.
-The application server follows the active-active architecture, and runs on the *first five nodes* in a cluster by default.
-Five is the maximum number of application servers per cluster.
-If one server falls offline, you can use the other servers without any loss of functionality.
+The application server follows the active-active architecture, and runs on the *first three nodes* in a cluster by default.
+
+NOTE: _**Three ** is the maximum number of application servers per cluster.
```
victorleeTG: We don't usually put notes in italics.  The NOTE box is deemed sufficient emphasis.

modules/cluster-and-ha-management/pages/ha-for-application-server.adoc
```diff
@@ -3,11 +3,13 @@
...
+If one server falls offline, you can use the other servers without any loss of functionality._
```
victorleeTG, line 9: This says, "you [the user] can use", seemingly putting the user in the driver's seat. However, an ideal system would have automatic failover, that is, the remaining servers that are still working will pick up the load from the one that is unavailable. The user would not have to lift a finger.
Please ask a developer if there is automatic failover, or does the user truly need to do something to get the system to work again. From a writing standpoint, which is the proper subject of the clause: "you" or "the remaining servers"?

victorleeTG (changes requested):

modules/cluster-and-ha-management/pages/ha-for-application-server.adoc
JosephNewmanTG: I asked the engineer this question on the ticket. If I don't get a response back I'll send them a message on slack. https://graphsql.atlassian.net/browse/DOC-1862?focusedCommentId=130829 

The only change I made was from five nodes to three nodes and the NOTE block as, I felt that paragraph needed to be broken up with something. I see your point, my impression with the DOC, was that it falls on the user, until they set up load-balancing, which is what the main chunk of the page is about. We will see though what the engineer says. Its a good point to highlight!

modules/cluster-and-ha-management/pages/ha-for-application-server.adoc
JosephNewmanTG: Good point! I removed the italics, with the last commit.

PR #334, opened 2023-09-29
Update to Cluster Restore

JosephNewmanTG:
Associated Task: https://graphsql.atlassian.net/browse/DOC-1830

This pull request updates the Cross Cluster Restore user process.

### DIFF
--- a/modules/backup-and-restore/pages/backup-cluster.adoc
+++ b/modules/backup-and-restore/pages/backup-cluster.adoc
@@ -39,8 +39,14 @@ $ gadmin backup create weekly
 
 This creates a backup file named `weekly-<timestamp>` at the xref:configurations.adoc[backup location you configured].
 
+
 [#_metadata_backup_optional]
 === Metadata backup (Optional)
+
+IMPORTANT: Since v3.9.3, it is no longer required to create a backup for a backups metadata.
+A backup’s metadata is now stored as a part of a backup's files and will be searched for with the backups position.
+
+NOTE: If you are using a version previous to v3.9.3, please refer to the instructions below.
 
 This step is only necessary if you are planning for an event where you need to restore the backup in another cluster.
 
--- a/modules/backup-and-restore/pages/backup-cluster.adoc
+++ b/modules/backup-and-restore/pages/backup-cluster.adoc
@@ -43,10 +43,10 @@ This creates a backup file named `weekly-<timestamp>` at the xref:configurations
 [#_metadata_backup_optional]
 === Metadata backup (Optional)
 
-IMPORTANT: Since v3.9.3, it is no longer required to create a backup for a backups metadata.
+IMPORTANT: Since v3.9.2, it is no longer required to create a backup for a backups metadata.
 A backup’s metadata is now stored as a part of a backup's files and will be searched for with the backups position.
 
-NOTE: If you are using a version previous to v3.9.3, please refer to the instructions below.
+NOTE: If you are using a version previous to v3.9.2, please refer to the instructions below.
 
 This step is only necessary if you are planning for an event where you need to restore the backup in another cluster.
 
--- a/modules/backup-and-restore/pages/cross-cluster-backup.adoc
+++ b/modules/backup-and-restore/pages/cross-cluster-backup.adoc
@@ -1,6 +1,5 @@
 = Restore a Database Backup from Another Cluster
 :description:
-:sectnums:
 
 This page walks you through the process of restoring a cluster from a backup that was created from another database cluster.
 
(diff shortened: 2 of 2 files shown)


modules/backup-and-restore/pages/backup-cluster.adoc
```diff
@@ -39,9 +39,15 @@ $ gadmin backup create weekly
 
 This creates a backup file named `weekly-<timestamp>` at the xref:configurations.adoc[backup location you configured].
 
+
 [#_metadata_backup_optional]
 === Metadata backup (Optional)
 
+IMPORTANT: Since v3.9.3, it is no longer required to create a backup for a backups metadata.
```
runlilong: since v3.9.2.

modules/backup-and-restore/pages/backup-cluster.adoc
```diff
@@ -39,9 +39,15 @@ $ gadmin backup create weekly
...
+A backup’s metadata is now stored as a part of a backup's files and will be searched for with the backups position.
+
+NOTE: If you are using a version previous to v3.9.3, please refer to the instructions below.
```
runlilong: v3.9.2

modules/backup-and-restore/pages/cross-cluster-backup.adoc
```diff
@@ -7,6 +7,35 @@ This page walks you through the process of restoring a cluster from a backup tha
...
+For the S3 backup, the restore flow will try to find an s3 object:
+
+[source, console]
+--
+<backup_tag>_metadata
+--
+
+From the Bucket `System.Backup.S3.BucketName`.
+
+IMPORTANT: For all previous versions please follow the steps below.
```
runlilong, line 52: for the version 3.9.2+, we still need step 2.2 https://docs.tigergraph.com/tigergraph-server/current/backup-and-restore/cross-cluster-backup#_copy_data_backup_to_new_cluster.

FYI:
     
compared to the previous version, what we can simplify are
1. there is no need to export the backup's metadata manually since the backup files have contained it.
2. no need to specify the metadata for the restore command, since it uses the metadata, at the backup files position, as default.

but we still support the old ways.

modules/backup-and-restore/pages/cross-cluster-backup.adoc
```diff
@@ -7,6 +7,35 @@ This page walks you through the process of restoring a cluster from a backup tha
 Restoring is an off-line operation.
 Your cluster is offline during the restore process.
 
+NOTE: Since v3.9.3, it is no longer required to create a backup for a backups metadata or upload the metadata to a new cluster.
```
runlilong: v3.9.2

modules/backup-and-restore/pages/backup-cluster.adoc
JosephNewmanTG: Thank you. Changed this with latest commit.

modules/backup-and-restore/pages/backup-cluster.adoc
JosephNewmanTG: Thank you. Changed this with latest commit.

modules/backup-and-restore/pages/cross-cluster-backup.adoc
JosephNewmanTG: Thank you fro the clarifications. I appreciate you. Changed this with latest commit.

modules/backup-and-restore/pages/cross-cluster-backup.adoc
JosephNewmanTG: Thank you. Changed this with latest commit.

JosephNewmanTG:
Merging this PR and closing the ticket.

victorleeTG (commented):
1. Organization: The two Prerequisites sections are almost identical.  I would be inclined to only list them once at the beginning, and have one bullet point:
* [3.9.1 and earlier] You have the metadata backup that goes with the data backup.
2. Organization: The 3.9.2+ steps are not in the right order. You have `gadmin backup restore` first and then `Copy data backup to new cluster`.
3. The two sections (pre-3.9.2, post-3.9.2) have different heading levels.
4. The two sections (pre-3.9.2, post-3.9.2) have different wording and level of details for S3 , but they don't actually conflict. Do the two product versions have the same actual S3 behavior? If so, use the same (best) wording. If they have different behavior, then spell it out more clearly.
5. For both S3 and local backups, it is incomplete to say "the restore flow will try to find the metadata at...".  The problem is "try".
(a) What if it doesn't find it there? Then what?
(b) I'm guessed that a 3.9.2+ backup operation puts the metadata file at
<backup_folder>/metadata. Please check with the developer.
(c) If that's true, then what the developer was saying is that the `restore` operation looks for the metadata in the place that the `backup` operation put it. Duh.
(Either explain this is full as an FYI, or don't explain it at all, because the intent is that TG handles this for you so that the users doesn't need to worry about the metadata.)
(d) The whole point of the "Copy data backup to new cluster" section is that 
i. There are configuration parameters for the new cluster to specify where to write/read backup files. Set those parameters.
ii. Put the backup files there.
Put currently, this section does not say that in a straightforward way.
6. I'm pretty sure that there are no differences for the non-metadata part of backup/restore, so that the whole "Copy data backup to new cluster" can/should read the same for both versions.

PR #341, opened 2023-10-12
CPU usage based query routing

JosephNewmanTG:
Associated Task: https://graphsql.atlassian.net/browse/DOC-1828

- Added enable instructions on workload management
- Added config commands configure parameters table

**additionally, this PR merges recent 3.9.2-1 changes into 3.9.3 staging branch

### DIFF
--- a/modules/system-management/pages/workload-management.adoc
+++ b/modules/system-management/pages/workload-management.adoc
@@ -42,18 +42,6 @@ $ gadmin confit set RESTPP.WorkLoadManager.MaxDelayQueueSize 20
 
 You must xref:manage-services.adoc#_start_stop_or_restart_a_service[restart the RESTPP service] for the change to take effect.
 
-== Specify replica to run query on
-On a distributed cluster, you can specify on which replica you want a query to be run through the xref:tigergraph-server:API:built-in-endpoints.adoc#_run_an_installed_query_post[Run Query REST endpoint].
-
-For example, to run the query on the primary cluster, use the `GSQL-REPLICA` header when running a query and set its value to 1:
-
-.Specify that the query run on the primary cluster
-[source.wrap,bash]
-----
-curl -X POST -H "GSQL-REPLICA: 1" -d '{"p":{"id":"Tom","type":"person"}}'
-"http://localhost:9000/query/social/hello"
-----
-
 == Specify number of threads used by a query
 You can specify the limit of the number of threads that can be used by one query through the xref:tigergraph-server:API:built-in-endpoints.adoc#_run_an_installed_query_post[Run Query REST endpoint].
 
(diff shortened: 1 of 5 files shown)


modules/system-management/pages/workload-management.adoc
```diff
@@ -64,3 +64,30 @@ For example, to specify a limit of four threads that can be used by a query, use
 ----
 curl -X POST -H "GSQL-THREAD-LIMIT: 4" -d '{"p":{"id":"Tom","type":"person"}}' "http://localhost:9000/query/social/hello"
 ----
+
+== Replica/Partition routing based on CPU Usage
```
victorleeTG: This new feature is thematically related to an existing feature "Specify Replica to Run Query On" and so should be presented in connection with it.  I would present Query Routing Schemas (round-robin and CPU load aware) first, and then Specify Replica (with the GSQL-REPLICA header) next.

Reason: Round-Robin is the default mode and so should come first. CPU Load Aware is an alternative to Round-Robin. The GSQL-Replica header is an override of any of these modes.

victorleeTG:
```suggestion
== Query Routing Schemes
```

modules/system-management/pages/workload-management.adoc
```diff
@@ -64,3 +64,30 @@ For example, to specify a limit of four threads that can be used by a query, use
...
+
+REST++ allows users to route queries based on CPU usage. Determined by a user configurable threshold (default set at 50%).
+This is so that queries don’t overwhelm nodes that are already under a heavy load.
```
victorleeTG:
```suggestion
In a distributed or replicated cluster, REST++ automatically routes queries to different GPEs, in order to spread the workload.

NOTE: If GSQL-REPLICA header is used when invoking a query, this header overrides the routing schema for that query.

=== Round Robin routing
 The default query routing scheme is round-robin. The first query is managed by GPE 0, the next query by GPE 1, and so on. After the last GPE, the cycle returns to GPE 0.
 
```

modules/system-management/pages/workload-management.adoc
```diff
@@ -64,3 +64,30 @@ For example, to specify a limit of four threads that can be used by a query, use
...
+If no GPE satisfies the CPU threshold condition, REST++ falls back to the default behavior (a round-robin selection from the full list of GPEs).
+
+NOTE: If GSQL-REPLICA header is used when invoking a query, this header would be used instead and ignore this routing logic.
+
+.Enable CPULoadAware routing
+[source, console]
+gadmin config entry RESTPP.QueryRouting.Mode
+RESTPP.QueryRouting.Mode [ 1 ]
+
+NOTE: Query routing mode has two options: 0=RoundRobin and 1=CPULoadAware.
```
victorleeTG:
```suggestion
Version 3.9.3 adds a system configuration parameter `RESTPP.CPULoadAware.Mode` to enable system administrators to select other query routing schemas:

* Mode = 0 (default): Round-Robin
* Mode = 1: CPU Load Aware 

=== CPU Load Aware Query Routing
When enabled, REST++ tries to direct incoming queries to the GPEs that are currently less busy.
Specifically, the system periodically polls CPU usage data to find a GPE whose CPU usage percentage is below `RESTPP.QueryRouting.TargetSelectionCPUThreshold` (default 50).

If no GPE satisfies the CPU threshold condition, REST++ falls back to the default behavior (round-robin selection).

```

victorleeTG: Avoid using NOTEs to highlight basic facts. Instead, organize and phrase your presentation so that the reader will naturally notice the basic facts.

modules/system-management/pages/workload-management.adoc
victorleeTG: Remember to point out when a feature is only available starting from 3.9.3. Please double check your other doc tickets.

modules/system-management/pages/workload-management.adoc
```diff
@@ -64,3 +64,30 @@ For example, to specify a limit of four threads that can be used by a query, use
...
+
+You can also then set the CPU usage threshold for CPULoadAware query routing mode.
+
+.Change CPU threshold
+[source, console]
+gadmin config entry RESTPP.QueryRouting.TargetSelectionCPUThreshold
+RESTPP.QueryRouting.TargetSelectionCPUThreshold [ 40 ]
+New: 40
+
+NOTE: Default CPU threshold is set at 50.
```
victorleeTG:
```suggestion
.Example: Change CPU Load Threshold and Enable CPU Load Aware routing
[source, console]
$ gadmin config entry RESTPP.QueryRouting.TargetSelectionCPUThreshold 40
$ gadmin config entry RESTPP.QueryRouting.Mode 1
```

victorleeTG (changes requested):
I ended up proposing a rewrite of all the information. It wasn't bad or wrong; I just wanted it to be better.  Happy to walk through this with you.

This comes out from stepping back and seeing how would you present all of the options to a new user, and to make sure users on different versions know what is/isn't available to them.


Also, as we discussed before, in the future, do not combine unrelated changes into the same PR.  The goal is each "feature" = 1 DOC ticket = 1 PR.

modules/system-management/pages/workload-management.adoc
JosephNewmanTG: This makes a lot of sense. Reordered the features in the latest commit.

JosephNewmanTG: updated this with latest commit.

modules/system-management/pages/workload-management.adoc
JosephNewmanTG: Updated this with latest commit.

modules/system-management/pages/workload-management.adoc
JosephNewmanTG: updated this with latest commit.

modules/system-management/pages/workload-management.adoc
JosephNewmanTG: updated this with latest commit.

modules/system-management/pages/workload-management.adoc
JosephNewmanTG: Good point. I'll keep this in mind.

JosephNewmanTG:
All changes have been made as discussed in our meeting on 10/13.

PR #354, opened 2023-10-26
K8's removing pages and links

JosephNewmanTG:
Associated task: https://graphsql.atlassian.net/browse/DOC-1846

Removed the pages that the GitHub Documents replaces:
- Changed there file names to .txt to remove them remove them from Antora without totally deleting them.
- Found place that were links to these pages and replace them with links to there equivalent in the GitHub Repo or commented them out or removed them.
- commented out these files from the nav.doc
- removed Kubernetes Classic links from 3.9.3, still available in previous versions.


modules/kubernetes/pages/index.adoc
```diff
@@ -23,15 +23,15 @@ This allows you to manage a TigerGraph cluster in a more natural and intuitive w
 
 Previous versions of TigerGraph used a different implementation of Kubernetes. This is still functional, but no longer supported. For long-term support and enhancement, use the Operator instead.
```
victorleeTG:
```suggestion
Previous versions of TigerGraph used a different implementation of Kubernetes.
The older versions may be still functional, but no longer supported. For long-term support and enhancement, use the Operator instead.
```
Ambiguous what is "this".
Also, because this changeover took place a while ago now, I'm not positive that the older versions still work with the newer database.

modules/kubernetes/pages/k8s-operator/index.adoc
```diff
@@ -7,35 +7,32 @@ It enables you to automate operations such as the creation, status checking, res
 
 IMPORTANT: Kubernetes Operator support is currently a Preview Feature. Preview Features give users an early look at future production-level features. Preview Features should not be used for production deployments.
 
-NOTE: Additionally, we have provided a public GitHub repo (https://github.com/tigergraph/ecosys/tree/master/k8s) to our preview documentation of the Kubernetes Operator. These docs are frequently updated to support operation, troubleshooting, and all other relevant use cases.
+NOTE: Additionally, we have provided a public xref:https://github.com/tigergraph/ecosys/tree/master/k8s[GitHub Repo] to our preview documentation of the Kubernetes Operator. These docs are frequently updated to support operation, troubleshooting, and all other relevant use cases.
```
victorleeTG: Should we be saying "additionally"?  Is GitHub the only documentation they have for now?

victorleeTG (changes requested):

modules/kubernetes/pages/index.adoc
JosephNewmanTG: Thank you. this was changed in the latest commit.

modules/kubernetes/pages/k8s-operator/index.adoc
JosephNewmanTG: Good question. I changed this with the latest commit.

PR #373, opened 2023-11-07
3.5_removing Avro from Kafka loader supported files

JosephNewmanTG:
Associated task: https://graphsql.atlassian.net/browse/DOC-1899

- Removing AVRO from Kafka Supported files 3.5

### DIFF
--- a/modules/data-loading/pages/kafka-loader/index.adoc
+++ b/modules/data-loading/pages/kafka-loader/index.adoc
@@ -20,7 +20,6 @@ If the loader is interrupted, you have to resume the loading job manually by run
 This resumes the loading job where it was interrupted.
 
 == Supported file formats
-* Avro
 * CSV
 * JSON
 


victorleeTG (approved):

modules/data-loading/pages/kafka-loader/index.adoc
```diff
@@ -20,7 +20,6 @@ If the loader is interrupted, you have to resume the loading job manually by run
 This resumes the loading job where it was interrupted.
 
 == Supported file formats
-* Avro
```
victorleeTG, line 23: Keep Avro, but add a note with a hyperlink that says something like
"Avro (when sourced from an external Kafka cluster using MirrorMaker)"

victorleeTG (dismissed):
As discussed, I think we should leave

victorleeTG (approved):

PR #374, opened 2023-11-07
3.6_removing Avro from Kafka loader supported files

JosephNewmanTG:
Associated task: https://graphsql.atlassian.net/browse/DOC-1899

- removing Avro from Kafka loader supported files 3.6

### DIFF
--- a/modules/data-loading/pages/kafka-loader/index.adoc
+++ b/modules/data-loading/pages/kafka-loader/index.adoc
@@ -31,6 +31,7 @@ A resumed job picks up from where it was stopped before and will not consume mes
 == Supported file formats
 * CSV
 * JSON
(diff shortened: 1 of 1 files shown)


modules/data-loading/pages/kafka-loader/index.adoc
```diff
@@ -29,7 +29,6 @@ A resumed job picks up from where it was stopped before and will not consume mes
 ====
 
 == Supported file formats
-* Avro
```
victorleeTG, line 32: Keep Avro, but add a note with a hyperlink that says something like
"Avro (when sourced from an external Kafka cluster using MirrorMaker)"

victorleeTG (changes requested):

modules/data-loading/pages/kafka-loader/index.adoc
JosephNewmanTG: Thank you I have added.

PR #390, opened 2023-11-22
Adding HA overview page

JosephNewmanTG:
This PR addresses this issue of Benefits and Limitations in TigerGraph High Availability Documentation. Please see this doc for more details: https://docs.google.com/document/d/1V0T9ULS4uPffgUHYfBflW4yGLWI2Xzg3VUZq2FeONek/edit?usp=sharing

It is also associated with this ticket: https://graphsql.atlassian.net/browse/DOC-1917

- Adding a HA overview page
- Navigation adjustments to any current HA pages

### DIFF
--- a/modules/cluster-and-ha-management/pages/ha-overview.adoc
+++ b/modules/cluster-and-ha-management/pages/ha-overview.adoc
@@ -2,27 +2,34 @@
 = High Availability Overview
 :description: Overview of High Availability functionality and supported features.
 
-A TigerGraph system with High Availability (HA) is a cluster of server machines which uses replication.
-Allowing for the database to continue to run with minimal impact to operations until the offline nodes are restored or removed permanently.
+A TigerGraph system with High Availability (HA) is a cluster of server nodes which uses replication to provide continuous operation of most services in the event that one or more nodes is offline.
 
 For example, an application's predefined queries will continue to run and file-based loading will continue with files resident at any of the working nodes.
 
-Additionally, incorporating HA can yield a number of other system benefits including:
+Incorporating HA can yield a number of other system benefits, such as:
 
-* Distribution of workloads across all primary and secondary nodes.
-* Data load operations to be distributed across all nodes.
-* Individual nodes to fail without impacting workloads.
-* Individual nodes can be re-paved or replaced without impacting workloads.
+* The query workload is distributed across all replicas.
+* Data loading operations are distributed across all nodes.
+* Individual nodes can fail without impacting query workloads.
+* Individual nodes can be re-paved without impacting query workloads.
 
-IMPORTANT: Considerations should be placed on the following instances when a node is offline:
+NOTE: Re-pavement of a node, is the process that takes a node offline intentionally for maintenance or updates and then is brought back online.
 
-* A file-based data load can only be loaded when the node is back online.
-* A data partition slated for a kafka-based data load, can resume once the node is back online.
-* New query installation will only resume once a node is back online; however, interpreted queries will continue as before.
-* A schema changes will return an error until the node is back online.
-* Backup and export operations continue as normal once failed nodes are back online or removed.
+== HA Considerations
 
-NOTE: Workarounds to the above instances can be archived by the xref:_removal_of_failed_nodes[removal of the failing nodes].
+TigerGraph HA provides continuous operation of some but not all services.
+Please note the following exceptions and consider your options for taking additional steps to maintain continuous operation or to restore service as quickly as possible.
+
+If an HA system is operating with a failed node, unless the system is reconfigured to xref:tigergraph-server:cluster-and-ha-management:crr-index.adoc#_exclusions[exclude that node], the following services are limited or unavailable:
+
+* A file-based data load cannot be loaded.
+* A data partition slated for a kafka-based data load cannot be loaded.
+* New queries cannot be installed.
+* Schema changes are not allowed.
+* Backup and export operations are not available and will be rejected.
+
+NOTE: As a workaround, full operation can be restored temporarily by the xref:_removal_of_failed_nodes[removal of the failing nodes].
+For example, a 5 x 2 cluster with one node removed would become a 5 x 4 cluster, with only partial replication.
 
 == High Availability Cluster Configuration
 
@@ -39,7 +46,7 @@ Key factors include:
 High Availability (HA) support for xref:tigergraph-server:cluster-and-ha-management:ha-for-gsql-server.adoc[GSQL Server] ensures robustness by allowing components to take over in case of unavailability.
 Incorporating a primary GSQL server and standby servers, collectively referred to as leaders and followers, offering HA for client connections.
 
-NOTE: It is recommended to couple this with a client-side retry logic, so failing user query will complete once the system adjusts to accommodate the failed node (typically up to 30 seconds).
+NOTE: It is recommended to couple this with a client-side retry logic, so the failing user query will complete once the system adjusts to accommodate the failed node (typically up to 30 seconds).
 
 == High Availability Support for Application Server
 
--- a/modules/cluster-and-ha-management/pages/ha-overview.adoc
+++ b/modules/cluster-and-ha-management/pages/ha-overview.adoc
@@ -1,77 +1,53 @@
-:toc:
-= High Availability Overview
+= High Availability (HA) Overview
 :description: Overview of High Availability functionality and supported features.
 
 A TigerGraph system with High Availability (HA) is a cluster of server nodes which uses replication to provide continuous operation of most services in the event that one or more nodes is offline.
 
 For example, an application's predefined queries will continue to run and file-based loading will continue with files resident at any of the working nodes.
 
-Incorporating HA can yield a number of other system benefits, such as:
-
+.Incorporating HA can yield a number of other system benefits, such as:
 * The query workload is distributed across all replicas.
+** This includes replicas used for long-running queries or with the xref:tigergraph-server:API:built-in-endpoints.adoc#_headers[GSQL-REPLICA] header.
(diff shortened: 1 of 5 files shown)


modules/cluster-and-ha-management/pages/ha-overview.adoc
```diff
@@ -0,0 +1,70 @@
+:toc:
+= High Availability Overview
+:description: Overview of High Availability functionality and supported features.
+
+A TigerGraph system with High Availability (HA) is a cluster of server machines which uses replication.
+Allowing for the database to continue to run with minimal impact to operations until the offline nodes are restored or removed permanently.
+
+For example, an application's predefined queries will continue to run and file-based loading will continue with files resident at any of the working nodes.
+
+Additionally, incorporating HA can yield a number of other system benefits including:
```
victorleeTG: While this phrasing preceding a list is commonly used, strictly speaking it is grammatical incorrect. The rule says that the colon should be preceded by a complete sentence. Here are some examples of a correctly used colon:
"Here are some examples of a correctly used colon:"
 "Incorporating HA can yield a number of other benefits, including the following:"
 "Incorporating HA can yield a number of other system benefits:"

modules/cluster-and-ha-management/pages/ha-overview.adoc
```diff
@@ -0,0 +1,70 @@
...
+
+* Distribution of workloads across all primary and secondary nodes.
```
victorleeTG:
```suggestion
* Distribution of workloads among the multiple replicas.
```
Our version of replication does not have a primary or secondary. All clusters are equal.

victorleeTG: The 4 items in this list are do not use similar sentence structure. Not good writing.
Item 1: no verb at all, just a noun phrase
Item 2: infinitive verb: "to be distributed"
Item 3: infinitive verb: "to fail"
Item 4: present tense, with modal auxiliary verb for possibility: "can be re-paved"

Because the list is abilities (which means a verb is important), I would suggest using complete sentences, such as, "The query workload is distributed across all replicas."

modules/cluster-and-ha-management/pages/ha-overview.adoc
```diff
@@ -0,0 +1,70 @@
...
+* Data load operations to be distributed across all nodes.
+* Individual nodes to fail without impacting workloads.
```
victorleeTG: Recall my question/comment from the meeting: What type of operations are included in "workload"?
We mean queries, but that is not obvious to a reader. If we mean queries, then we should say queries, as in my example for item 1.

modules/cluster-and-ha-management/pages/ha-overview.adoc
```diff
@@ -0,0 +1,70 @@
...
+* Individual nodes can be re-paved or replaced without impacting workloads.
```
victorleeTG: Add a footnote to explain the term "re-pave."

modules/cluster-and-ha-management/pages/ha-overview.adoc
```diff
@@ -0,0 +1,70 @@
...
+
+IMPORTANT: Considerations should be placed on the following instances when a node is offline:
```
victorleeTG: I think a subsection is better than having in an admonition block. Here is a possible heading:

## HA Considerations
TigerGraph HA provides continuous operation of some but not all services.
Please note the following exceptions and consider your options for taking additional steps to maintain continuous operation or to restore service as quickly as possible.

If an HA system is operating with a failed node, unless the system is **reconfigured to exclude that node**[hyperlink], the following services are limited or unavailable:

modules/cluster-and-ha-management/pages/ha-overview.adoc
```diff
@@ -0,0 +1,70 @@
...
+
+* A file-based data load can only be loaded when the node is back online.
```
victorleeTG: This is not worded clearly. It sounds like file-based loading simply doesn't work (which contracts the earlier claim about what does continue to work).  Explain what doesn't. The part "until the node is back online" is mostly likely unncessary and just adding complication to the sentence. Just explain what doesn't work when a node is down.

victorleeTG: Rather than having each list item say "until the node is back online", it would clearer to say in one sentence, after this list, what happens when the node is restored or replaced? Does everything work again? What about jobs that were running when the node went down?

modules/cluster-and-ha-management/pages/ha-overview.adoc
```diff
@@ -0,0 +1,70 @@
...
+* A data partition slated for a kafka-based data load, can resume once the node is back online.
+* New query installation will only resume once a node is back online; however, interpreted queries will continue as before.
+* A schema changes will return an error until the node is back online.
+* Backup and export operations continue as normal once failed nodes are back online or removed.
```
victorleeTG: Improve all of these.  
Example:
* Backup and export operations are not available and will be rejected.

modules/cluster-and-ha-management/pages/ha-overview.adoc
```diff
@@ -0,0 +1,70 @@
...
+
+NOTE: Workarounds to the above instances can be archived by the xref:_removal_of_failed_nodes[removal of the failing nodes].
```
victorleeTG:
```suggestion
NOTE: As a workaround, full operation can be restored temporarily by xref:_removal_of_failed_nodes[removal of the failing nodes].
For example, a 5 x 2 cluster with one node removed would become a 5 + 4 cluster, with only partial replication. 
```

modules/cluster-and-ha-management/pages/ha-overview.adoc
```diff
@@ -0,0 +1,70 @@
...
+* GCP External HTTP(s) Load Balancer
+
+== Cluster Commands
+
+xref:tigergraph-server:cluster-and-ha-management:cluster-commands.adoc[Advanced Linux commands] that simplify platform operations can performed by the platform owner during debugging on HA clusters.
+
+== Removal of Failed Nodes
+
+HA cluster configuration allows for the xref:tigergraph-server:cluster-and-ha-management:remove-failed-node.adoc[removal of failed nodes] without data loss, facilitating cluster maintenance and recovery.
+This feature is especially useful when addressing individual node failures or handling multiple concurrent node failures in a prepared environment, covering:
```
victorleeTG: Not here but on the Removal of Failed Nodes page, I think we can improve the explanation with an illustrative example.

modules/cluster-and-ha-management/pages/ha-overview.adoc
```diff
@@ -0,0 +1,70 @@
+:toc:
+= High Availability Overview
+:description: Overview of High Availability functionality and supported features.
+
+A TigerGraph system with High Availability (HA) is a cluster of server machines which uses replication.
+Allowing for the database to continue to run with minimal impact to operations until the offline nodes are restored or removed permanently.
```
victorleeTG:
```suggestion
A TigerGraph system with High Availability (HA) is a cluster of server nodes which uses replication to provide continuous operation of most services in the event that one or more nodes is offline. 
```
Your split into two sentences is not grammatical. 
I think "continuous operation of most services" or "many services" is a good way to say what we do.

victorleeTG (changes requested):

modules/cluster-and-ha-management/pages/ha-overview.adoc
JosephNewmanTG: Thank you fixed this will latest commit.

modules/cluster-and-ha-management/pages/ha-overview.adoc
JosephNewmanTG: Thank you addressed this with latest commit.

modules/cluster-and-ha-management/pages/ha-overview.adoc
JosephNewmanTG: Thank you added footnote.

modules/cluster-and-ha-management/pages/ha-overview.adoc
JosephNewmanTG: Thank you. I added this will latest commit.

modules/cluster-and-ha-management/pages/ha-overview.adoc
JosephNewmanTG: Thank you fixed this will latest commit.

modules/cluster-and-ha-management/pages/ha-overview.adoc
JosephNewmanTG: Fixed these with latest commit.

modules/cluster-and-ha-management/pages/ha-overview.adoc
JosephNewmanTG: Added this with latest commit.

modules/cluster-and-ha-management/pages/ha-overview.adoc
JosephNewmanTG: thank you fixed this with latest commit.

modules/cluster-and-ha-management/pages/ha-overview.adoc
JosephNewmanTG: This is a good idea. Let's make it a separate ticket and PR though. This main draft of the overview page is what needs to be considered at this immediate time.

modules/cluster-and-ha-management/pages/ha-overview.adoc
JosephNewmanTG: thank you. addressed this with latest commit.

PR #453, opened 2024-07-03
DOC-2197 docs(backup): add s3 backup roleARN config to backup doc;

zejunlitg:
ticket link: https://graphsql.atlassian.net/browse/DOC-2197

In `configurations.adoc`:
Screenshot 2024-08-26 at 6 35 39 PM

Added sections:
Screenshot 2024-08-26 at 6 35 54 PM

In `configuration-parameters.adoc`:
Screenshot 2024-08-26 at 6 41 14 PM

### DIFF
--- a/modules/backup-and-restore/pages/configurations.adoc
+++ b/modules/backup-and-restore/pages/configurations.adoc
@@ -32,6 +32,11 @@ Required if backup is to be stored on S3.
 *NOTE*: If setting this in interactive mode, store the key in a file and provide the path to the file, e.g., `@/tmp/test_secret`.
 |`+nan+`
 
+|System.Backup.S3.RoleARN |The AWS role for accessing s3 bucket, its use is prioritized over the combo of access key id and secret access key in accessing s3. Note this is only for AWS S3, and that using role ARN assumes access key id and secret access key are already configured with AWS CLI
+
+*NOTE*: This is only for AWS S3, and that using role ARN assumes access key id, secret access key and region are already configured with AWS CLI `aws configure`.
+|`+nan+`
+
 |System.Backup.S3.BucketName |Name of the S3 bucket to store backup files.
 Required if backup is to be stored on S3.|null
 
@@ -139,6 +144,12 @@ gadmin config set "System.Backup.S3.AWSAccessKeyID" "<access-key-id>"
 [console,]
 ----
 gadmin config set "System.Backup.S3.AWSSecretAccessKey" "<secret-access-key>"
+----
++
+.Alternatively, instead of using `AccessKeyID` and `SecretAccessKey`, you may use AWS Role ARN for the authentication
+[console,]
+----
+gadmin config set "System.Backup.S3.RoleARN" "<aws-role-arn>"
 ----
 +
 .Apply the new parameter values
--- a/modules/backup-and-restore/pages/configurations.adoc
+++ b/modules/backup-and-restore/pages/configurations.adoc
@@ -32,9 +32,9 @@ Required if backup is to be stored on S3.
 *NOTE*: If setting this in interactive mode, store the key in a file and provide the path to the file, e.g., `@/tmp/test_secret`.
 |`+nan+`
 
-|System.Backup.S3.RoleARN |The AWS role for accessing s3 bucket, its use is prioritized over the combo of access key id and secret access key in accessing s3. Note this is only for AWS S3, and that using role ARN assumes access key id and secret access key are already configured with AWS CLI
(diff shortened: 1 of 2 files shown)


modules/reference/pages/configuration-parameters.adoc
```diff
@@ -1243,6 +1243,8 @@ backup |`nan`
 |System.Backup.S3.AWSSecretAccessKey |The secret access key for s3
 bucket |`nan`
 
+|System.Backup.S3.RoleARN |The AWS role for accessing S3 bucket, its use is prioritized over the combo of access key ID and secret access key in accessing S3. Note this is only for AWS S3, and that using role ARN assumes access key ID and secret access key are already configured with AWS CLI. |`nan`
```
chengjie-qin: Do we need to give any examples or it should be straightforward enough for people who uses role arn?

modules/backup-and-restore/pages/configurations.adoc
```diff
@@ -32,6 +32,11 @@ Required if backup is to be stored on S3.
 *NOTE*: If setting this in interactive mode, store the key in a file and provide the path to the file, e.g., `@/tmp/test_secret`.
 |`+nan+`
 
+|System.Backup.S3.RoleARN |The AWS role for accessing s3 bucket, its use is prioritized over the combo of access key id and secret access key in accessing s3. Note this is only for AWS S3, and that using role ARN assumes access key id and secret access key are already configured with AWS CLI
```
runlilong: "role ARN assumes access key id and secret access key" is not a must. we only need the cluster nodes to be allowed to assume that role.  there can be several ways,  refer here https://aws.amazon.com/blogs/security/how-to-use-trust-policies-with-iam-roles/

Sorry, I didn't catch this in the PR review. If this is confirmed, we also need to update the description of `System.Backup.S3.RoleARN` in the code.

so, how about this one:
 "Note this is only for AWS S3, and it's crucial to ensure that your nodes are trusted and permitted to assume this RoleARN. To verify this, you can use the `aws sts assume-role` command."

feel free to make it better.

modules/reference/pages/configuration-parameters.adoc
runlilong: The usage of RoleARN is highly dependent on the node environment. Providing detailed configuration instructions might be too complex for our purposes, as it primarily pertains to AWS S3. How about we suggest users verify it first using the `aws sts assume-role` command?

modules/backup-and-restore/pages/configurations.adoc
zejunlitg: I've changed the description based on our discussion offline.

modules/reference/pages/configuration-parameters.adoc
zejunlitg: @chengjie-qin I think customers who ask about this should already be aware of this. If we were to promote what we support, we can always explain when we do the promotion. The description will make sure people who know about it knows how to make it work with TigerGraph.

zejunlitg: @runlilong I've updated both places with same description to make it consistent per our discussion.

runlilong (approved):

modules/reference/pages/configuration-parameters.adoc
arun-TG-PM: @zejunlitg @runlilong  - I definitely see value in providing a simple example with a specific persona such as DBA.

Example: Configuring S3 Backup with AWS Role ARN

Let’s walk through a scenario where Andrea, a Database Admininstrator (DBA) at a mid-sized company, is responsible for setting up automated backups of her TigerGraph cluster to an AWS S3 bucket. Instead of using long-lived credentials (AccessKeyID and SecretAccessKey), she decides to use an AWS Role ARN for enhanced security and easier management.

Step 1: Create an AWS IAM Role
Explain
Step 2: Assign the Role ARN
Explain
Step 3: Apply configuration 
Explain
Step 4: Verify the Backup configuration
Explain

chengjie-qin (approved):

modules/reference/pages/configuration-parameters.adoc
zejunlitg: @arun-TG-PM I can add the example, but not sure where would be a suitable place to add that as these 2 places in the PR are meant to be concise?
cc @victorleeTG

zejunlitg: just realized there is example already in `configurations.adoc`, to help understand I'm using `arn:aws:iam::account:role/role-name-with-path` from AWS role ARN doc: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html#identifiers-arns

arun-TG-PM (approved):

PR #479, opened 2024-07-23
[DOC-2240] fix(ddl): Add header option for ddl endpoint;

JunruZhu:


modules/API/pages/built-in-endpoints.adoc
```diff
@@ -1527,6 +1527,10 @@ curl -X POST --data-binary @./work_at.csv "http://localhost:9000/ddl/poc_graph?t
 | Yes
 | File variable name or file path for the file containing the data
 
+| `header`
+| No
+| Whether to skip loading the header line of the file or not. The input should be true or false. The default header is false.
```
victorleeTG:
```suggestion
| Whether to skip loading the header line of the file or not. The value should be either true or false. The default value is false.
```

victorleeTG (changes requested):

PR #504, opened 2024-08-29
DOC-2229-fine-grained-query-priv

victorleeTG:


victorleeTG:
* Add the new page fine-grained-query-privileges to the navigation menu
* Remove the old page acl-management from the navigation menu
* Make some writing style and wording improvements to the fine-grained-query-privileges page.

PR #561, opened 2024-10-17
DOC-2347 Docker: Add note on using different container and volume names (all ver)

szilardbaranytg:
… in case of having multiple TG deployments in the same Docker environment
(there is a typo: "specify a a different", please fix it)

### DIFF
--- a/modules/getting-started/pages/docker.adoc
+++ b/modules/getting-started/pages/docker.adoc
@@ -102,6 +102,9 @@ If you use Windows and have write permission issues with the above command,  try
 $ docker run -d -p 14022:22 -p 9000:9000 -p 14240:14240 --name tigergraph --ulimit nofile=1000000:1000000 -t tigergraph/tigergraph:latest
 ----
 
+[NOTE]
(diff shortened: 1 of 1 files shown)


modules/getting-started/pages/docker.adoc
```diff
@@ -102,6 +102,9 @@ If you use Windows and have write permission issues with the above command,  try
 $ docker run -d -p 14022:22 -p 9000:9000 -p 14240:14240 --name tigergraph --ulimit nofile=1000000:1000000 -t tigergraph/tigergraph:latest
 ----
 
+[NOTE]
+If you intend to have multiple versions or instances of the same version of TigerGraph in the same Docker environment, make sure that you specify a a different <3> contain name and <6> volume name to avoid collusion and accidental reuse of an earlier volume.
```
victorleeTG:
```suggestion
If you intend to have multiple instances of TigerGraph in the same Docker environment, whether they are the same version or not, we advise using unique <3> container names and <6> volume names for each one, to avoid accidental reuse or interference between volumes.
```

PR #576, opened 2024-10-23
Update README.adoc for link fix

deepkhajanchi:
Clicking on the link shows this url on address bar: https://docs.tigergraph.com/tigergraph-server/4.1/data-querying/readme#3.2@gsql-ref:querying:introduction-query.adoc, but it doesn't redirect to anywhere.


victorleeTG:
This problem was fixed in another PR.

PR #588, opened 2024-10-24
DOC-2031 add notation to show this is added in 3.11

victorleeTG:


modules/system-management/pages/change-data-capture/cdc-message-example.adoc
```diff
@@ -54,7 +54,7 @@ The current message is the second message in the third batch of the transaction.
 
 “*Transaction*: epoch time of transaction commit message.”
 
-¦ `graph`
+¦ `graph`[3.11+]
```
victorleeTG:
```suggestion
¦ `graph` [3.11+]
```

PR #639, opened 2024-12-02
DOC-2386

Tushar-TG-14:


modules/API/pages/upsert-rest.adoc
```diff
@@ -316,24 +316,70 @@ curl -X POST "localhost:14240/restpp/graph/Person_Movie" -d '
 
 == Upserting edges with discriminators
 
-Some edge types are defined with xref:gsql-ref:ddl-and-loading:defining-a-graph-schema.adoc#_discriminator[discriminators], which allow multiple instances of the same edge type between two vertices.
+Some edge types are defined with xref:gsql-ref:ddl-and-loading:defining-a-graph-schema.adoc#_discriminator[discriminators], which allow multiple instances of the same edge type between two vertices. Discriminators typically consist of one or more attributes that uniquely identify the edge.
```
victorleeTG:
```suggestion
Some edge types are defined with xref:gsql-ref:ddl-and-loading:defining-a-graph-schema.adoc#_discriminator[discriminators], which allow multiple instances of the same edge type between two vertices.
Discriminators typically consist of one or more attributes that uniquely identify the edge.
```
Put new sentences on a separate line.

modules/API/pages/upsert-rest.adoc
```diff
@@ -316,24 +316,70 @@ curl -X POST "localhost:14240/restpp/graph/Person_Movie" -d '
...
+
+[source.wrap,console]
+----
+'{"edges":{"Card":{"{\'1234-124-1245\': {\'Transacted_At\': {\'Merchant\': {\'TGT100\': {\'amount\': {\'value\': 99.95}, \'time\': {\'value\': 0}}}}}}":{"Transacted_At":{"Merchant":{"{\'TGT100\': {\'amount\': {\'value\': 99.95}, \'time\': {\'value\': 0}}}":{"amount": {"value": {"keylist": ["value"], "valuelist": [99.95]}}, "time": {"value": {"keylist": ["value"], "valuelist": [0]}}}}}}}}}}'
+----
+
+[NOTE]
+====
+Upserting edges with discriminators involves a non-standard process in TigerGraph’s REST API. 
+You will need to convert the edge data into a specific string format (using *keylist/valuelist* pairs) for successful upsertion.
```
victorleeTG: These notes seem a little vague to me. The user can see the indented JSON and the string, so at a high-level, they can see what to do.  The notes should point out specific rules that might not be obvious:

1. Because the whole data object is enclosed in double quotes, the quotes for the JSON key and value fields need to be converted from double quotes (") to escaped single quotes (\').
2. There is a 2nd point about how to deal with fields that have a list of set of values. The ticket has an example in the description. You need to either modify this example or add a 2nd example in order to show a list of set. 
'In order to upsert a list/set of values, the data is formatted with a set of key-value pairs with scalar values. A future version of TigerGraph will also support a single key-value pair having a set/list value.'

victorleeTG (changes requested):

modules/API/pages/upsert-rest.adoc
victorleeTG: Also, the data example (Transacted_At) uses a different schema than you schema example (Study_At). Let's make them match.

modules/API/pages/upsert-rest.adoc
```diff
@@ -314,26 +315,175 @@ curl -X POST "localhost:14240/restpp/graph/Person_Movie" -d '
...
-CREATE DIRECTED EDGE Study_At(From Person, To University, DISCRIMINATOR(class_year INT, class_month INT), major STRING))
+CREATE DIRECTED EDGE Liked (  
+  FROM User,  
+  TO User,  
+  DISCRIMINATOR(actionId STRING),  
+  weight FLOAT  
+);  
+----
+
+==== Old format (Single edge):
```
victorleeTG: We should avoid saying "new" and "old" unless we have clearly established in this context what historical event we are talking about.  You an explanation for why there are two formats. Also, I believe the developer added a comment recently that explained even in the "old" way, each edge of a multi-edge count in inserted one at a time, using standard JSON.

victorleeTG (approved):

PR #651, opened 2024-12-16
DOC-2420

Tushar-TG-14:


modules/release-notes/pages/index.adoc
```diff
@@ -209,6 +209,21 @@ AdminPortal UI for the xref:#finer_grain_query_privileges[fine-grained query pri
 
 == Fixed issues
 
+=== Fixed and Improved in 4.1.2
+
+==== Functionality
+* Fixed issue where query installation failed for single-node queries that initialize vertex set variables in conditional branches (e.g., `if-else` or `case-when` statements) (GLE-8908).
+* Fixed disk critical issue caused by the rebuilder getting stuck in a partitioned cluster after dropping vertex or edge attributes (CORE-4418).
+* Fixed issue with error codes in the log when a CDC message failed to deliver to external Kafka (CORE-4425).
```
victorleeTG: Most of the descriptions give some clue to the type of problem:
"query INSTALLATION FAILED"
"DISK [space] CRITICAL"
"QUERY COMPILATION ERROR"
but this one doesn't give much of a hint about the type of problem:
"Issue with error codes."  Guess: were we reporting the wrong codes?
When I look at the ticket https://graphsql.atlassian.net/browse/CORE-4425, it looks to me that the problem was that when a CDC message failed to be delivered (which is a serious error), it was not being recorded properly in the log UNLESS the log level was DEBUG.  (You should be familiar with the log levels https://docs.tigergraph.com/tigergraph-server/4.1/troubleshooting/log-files#_tigergraph_log_structure

modules/release-notes/pages/index.adoc
```diff
@@ -209,6 +209,21 @@ AdminPortal UI for the xref:#finer_grain_query_privileges[fine-grained query pri
...
+* Fixed query compilation error when iterating through a `foreach` statement to retrieve an `accum` type attribute in a query compiled with the `-single` or `distributed` keyword. Ensured operations between two different `GroupByAccums` hold the same data types and field names (GLE-9054).
+* Fixed restore failure when importing GSQL data due to excessive time taken by the GSQL warmup process (TP-6948).
+* Fixed an exception (`NoSuchElementException`) when removing all tokens (GLE-9022).
```
```suggestion
* Fixed an exception (`NoSuchElementException`) when removing all access tokens (GLE-9022).
```
Clarify the context of "token".

modules/release-notes/pages/index.adoc
```diff
@@ -209,6 +209,21 @@ AdminPortal UI for the xref:#finer_grain_query_privileges[fine-grained query pri
...
+
+==== Improvements
+* Improved the performance of GSQL queries containing delete statements intended for deleting all vertices of a given type (GLE-8913).
+* Added support for customizing pod labels and annotations in the `PodTemplateSpec` of the StatefulSet managing TigerGraph pods (TP-6901).
```
victorleeTG: "Added support for customizing" sounds like the user would need to do sometime to take advantage of this feature, which means some documentation should be required.  Looking at the eng ticket, I see
">> Add two new fields, PodInitLabels and PodInitAnnotations
I would ask the engineer if the user needs to do someone to use these fields.

If there is a functional change, we should have a DOC ticket. Currently, the engineering team is writing Kubernetes documentation in GitHub.
I did not find either of those fields in https://github.com/tigergraph/ecosys/tree/master/k8s

modules/release-notes/pages/index.adoc
```diff
@@ -209,6 +209,21 @@ AdminPortal UI for the xref:#finer_grain_query_privileges[fine-grained query pri
...
+* Resolved a CDC `UNKNOWN` ID issue occurring under high load in the GSE service (CORE-4444).
```
victorleeTG: This sounds like a Fixed Issue instead of an improvement.

victorleeTG (changes requested):

victorleeTG (approved):

PR #664, opened 2024-12-20
DOC-2420

Tushar-TG-14:


victorleeTG (approved):
@Tushar-TG-14 I actually would list this as a feature change. 
"Improvements" should qualitative issues, like performance improvement.

PR #670, opened 2025-01-07
DOC-2276-JWT token support

Tushar-TG-14:


modules/user-access/pages/jwt-token.adoc
```diff
@@ -407,6 +420,8 @@ When using a JWT token for authentication, please consider these scenarios and h
 
 ¦ Expiration ¦ If the token has expired, an error is thrown to indicate its expiration.
 
+¦ Unsupported Features ¦ GSQL endpoints do not support JWT tokens in these versions.
```
victorleeTG: @Tushar-TG-14  This table is to explain the types of error messages the function can return. I don't think  you can add to this table.

modules/user-access/pages/jwt-token.adoc
```diff
@@ -4,6 +4,14 @@ OpenID Connect (OIDC) token-based authentication in JSON web token (JWT) format,
 Helping to secure users and their data.
 
 == OIDC JWT Authentication in TigerGraph
+
+[NOTE]
+====
+* In versions 3.10 and 3.11, JWT token support is limited to RESTPP endpoints and only for third-party JWT tokens.
+* GSQL server endpoints do not support JWT tokens in these versions.
+* Internally generated JWT tokens are also not supported in these versions.
```
victorleeTG, line 12:
```suggestion
* In versions 3.10 and 3.11, JWT token support is limited to RESTPP endpoints and only for third-party JWT tokens.
* GSQL server endpoints do not support JWT tokens in these versions.
* Internally generated JWT tokens are also not supported in these versions.

Version 4.1.2 adds support for internally-generated JWT tokens and use with GSQL server endpoints, removing these limitations.
```
If we had written this when we should have (with the 3.10 release), we would not know for certain when we would remove these limitations. We would have to phrase it differently, e.g.
" The current version does not yet support <these features>. We plan to remove these limitations in a future release."

victorleeTG (changes requested):

PR #678, opened 2025-01-21
DOC-2374

anjaliIthapeTG:


modules/API/pages/built-in-endpoints.adoc
```diff
@@ -2867,6 +2867,27 @@ curl --fail -u <my_username>:<my_password> -X POST "http://localhost:14240/gsql/
...
+
+* GSQL Command
+
+[source.wrap,bash]
+----
+curl -s -H "Authorization: Bearer <USER TOKEN>" -H 'Workload-Queue: OLTP' 'localhost:14240/gsqlserver/interpreted_query' \
+-d 'INTERPRET QUERY -QUEUE OLTP () FOR GRAPH g { print 2; }'
+----
+
+[NOTE]:If both methods are used in a single request, the workload queue specified in the header and the GSQL command must match.
```
victorleeTG, line 2890: I know you just copied the text in the doc tickets description, but I see these are not good examples. Are there are some bigger problems.
(1) The two examples have one difference: the second one includes `-QUEUE OLTP` in the query body.
What is the other way? It's `-H 'Workload-Queue: OLTP'`. The -H option means "include this in the the REST request's header". This method was not called out clearly, but it actually the one thing this edit should be focused on.

victorleeTG (changes requested):

PR #696, opened 2025-02-05
DOC-2424-Added remaining fixes

Tushar-TG-14:


modules/release-notes/pages/index.adoc
```diff
@@ -177,6 +177,11 @@ See xref:3.10@tigergraph-server:user-access:rbac-row-policy/row-policy-privilege
 * Fixed an issue where schema changes could fail to occur during the loading process due to a lock not being removed (GLE-7572).
 * Fixed disk critical issue caused by the rebuilder getting stuck in a partitioned cluster after dropping vertex or edge attributes (CORE-4356).
 * Fixed an issue where failed backups in rare cases could cause the engine to pause for a long time, leading to query timeouts (CORE-4499).
+* Fixed issue where CDC EDGE messages with "UNKNOWN" ID were not removed when a query deleted a vertex and its associated edge (CORE-4491).
+* Resolved "UNKNOWN" ID issue in CDC messages under heavy load conditions on GSE (CORE-4443).
```
victorleeTG: I would like to know what "Resolved" means.  "UNKNOWN" is not a good thing, but it seems to be the result if the GSE is reset while CDC messages are being generated: https://docs.tigergraph.com/tigergraph-server/4.1/system-management/change-data-capture/cdc-overview#_when_gse_is_reset

Does this line mean we had an additional situation (heavy load) that could cause "UNKNOWN" ids but we have now fixed it?  Can we say "Fixed" instead of "resolved"?

victorleeTG: I want to point out that I knew nothing about UNKNOWN id in CDC until I read this, but then I searched the documentation to understand as much as I could before I formed my questions. When we write, we need to try to learn about the feature that we are describing so that we can judge if what we are saying makes sense or not.

modules/release-notes/pages/index.adoc
```diff
@@ -177,6 +177,11 @@ See xref:3.10@tigergraph-server:user-access:rbac-row-policy/row-policy-privilege
 * Fixed an issue where schema changes could fail to occur during the loading process due to a lock not being removed (GLE-7572).
 * Fixed disk critical issue caused by the rebuilder getting stuck in a partitioned cluster after dropping vertex or edge attributes (CORE-4356).
 * Fixed an issue where failed backups in rare cases could cause the engine to pause for a long time, leading to query timeouts (CORE-4499).
+* Fixed issue where CDC EDGE messages with "UNKNOWN" ID were not removed when a query deleted a vertex and its associated edge (CORE-4491).
```
victorleeTG, line 180: This one is puzzling.
 This seems to be saying that a 'query deleting a vertex and its associated edge'  SHOULD 'remove CDC EDGE messages with "UKNOWN" ID".'  The removal wasn't happening but now we've fixed it.
 
 But the SHOULD behavior is totally confusing.  Now, it might be related to this explanation of limitations:
 https://docs.tigergraph.com/tigergraph-server/4.1/system-management/change-data-capture/cdc-overview#_stuck_at_special_uid_unknown.
 But this is only talking about getting unstuck, not about changing any messages with "UNKNOWN".

victorleeTG (commented):
@Tushar-TG-14 
2 issues:
1. Ask the developer if it would be correct to say "Fixed" instead of "Resolved". If not, what should we say?  Resolved does not describe how something changed.

2. The other comment. The line might be "okay", but there is a larger issue that our CDC documentation does not tell users the recommended approach for dealing with "UNKNOWN" ids. I would not try to fix this sentence in this ticket. I would open a new ticket to address "In CDC, what does the TigerGraph product do, and what should users do, when an UNKNOWN id occurs?"

modules/release-notes/pages/index.adoc
Tushar-TG-14: @victorleeTG 
This is what engineer has replied:

It's a random issue when GSE in heavy load, not part of the GSE reset. Yes, we can say "Fixed" instead of "resolved".

modules/release-notes/pages/index.adoc
Tushar-TG-14: @victorleeTG 
This is what engineer has replied:

We noticed that when a query delete vertex and associated edge at same time, the EDGE message will have "UNKNOWN" id. It's too difficult to fix "UNKNOWN" id to correct id for such EDGE message. So we decided to simply remove such EDGE message, and let customer handle it, as in our limit "No CDC message for implicit edge deletion". So the term "should" can be confusing, because it's a decision made from implementation.

victorleeTG (approved):

PR #705, opened 2025-02-12
DOC-2405-Upsert Edge with Discriminator is a standard JSON Array

Tushar-TG-14:


victorleeTG (approved):
@Tushar-TG-14 Approved, but the section "JSON Format for Upserting Discriminated Edges" (single discriminated edge) is not needed at all.  A user who understands JSON will realize that the multi-edge syntax can be used to upsert a single single.  Please remove that section.

PR #721, opened 2025-02-26
DOC-2484-to-add-fixed-version-on-release-notes-document-webpages-3-11

anjaliIthapeTG:


modules/release-notes/pages/index.adoc
```diff
@@ -185,8 +185,10 @@ for more context-aware and interactive dashboard displays
 * Fixed situation where a query containing a `BREAK` or `CONTINUE` statements could produce incorrect results (GLE-7874).
 * Fixed regression problem with installing queries which create lists containing mixed types of numeric data (GLE-7928).
 * Fixed int64 value underflow error by explicitly type casting uint64 (CORE-4108).
+* Fixed issue where When using IMPORT ALL if a users schema size in the .zip file is exceedingly large, the import fails with an error message:Large catalog file key: /1/ReplicaList.json(GLE-6505).
```
```suggestion
* Fixed an issue so IMPORT ALL will no longer fail due to the schema size being very large (GLE-6505).
```
The fix is not making the error message go away.
The fix is that there is no longer a failure (I hope!).

PR #736, opened 2025-03-05
DOC-2070-CDC causes GPE crash [3.10]

Tushar-TG-14:


modules/system-management/pages/change-data-capture/cdc-message-example.adoc
```diff
@@ -259,6 +259,28 @@ For now, TigerGraph CDC Service will *NOT* produce CDC messages for such “impl
 ¦ Deletion of an edge
 |===
 
+In TigerGraph version 3.10, using the CDC feature with multi-edge types can cause a *GPE crash* when a `DELETE` operation is performed without specifying a discriminator.
```
victorleeTG, line 262: @Tushar-TG-14 Your description is very good. But its also inappropriate. I told you I didn't want a GPE crash mentioned in the feature documentation.  The crash should only be mentioned in the Known Issues in the Release Notes.

Think of it this way: A GPE crash is UNACCEPTABLE for production use.  If there is a way a user can avoid it, then they should. What would do:
In the CDC Limitations section (https://docs.tigergraph.com/tigergraph-server/3.10/system-management/change-data-capture/cdc-overview#_cdc_limitations):

[WARNING]
The current version of CDC is unstable when
1. The schema had edges with discriminators AND
2. A `DELETE` edges operation is performed without specifying a discriminator.
If CDC is being used, do not perform this type of unrestricted deletion.

victorleeTG (changes requested):

PR #740, opened 2025-03-07
[DOC-2558] docs(GSQLAPI): Add authentication header for GSQL example request;

ArthurXie-tg:
https://graphsql.atlassian.net/browse/DOC-2558

1. Emphasised 3 auth approaches at the begin of the doc page
2. Add authentication header to example requests (mix of user-password and jwt-token)

### DIFF
--- a/modules/API/pages/gsql-endpoints.adoc
+++ b/modules/API/pages/gsql-endpoints.adoc
@@ -16,6 +16,16 @@ You can refer to the xref:built-in-endpoints.adoc[REST endpoints for TigerGraph 
 
 NOTE: As of version 4.1.2, JWT tokens can be used for authentication with GSQL server REST endpoints.
 
+[NOTE]
+====
+For requests that are sent to the GSQL server, the sender needs to provide TigerGraph user credentials for the request to be accepted.
+User credentials can be provided in 3 ways:
+
+* `-u <user:password>`: pass the specified user name and password with curl option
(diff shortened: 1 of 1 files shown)


dadongwang-tg (approved):

victorleeTG (changes requested):
@ArthurXie-tg The page https://docs.tigergraph.com/tigergraph-server/4.1/api/ already explains the authentication requirement/header. 
Users should think and want to think of all endpoints as one group; not GSQL endpoints and other endpoints.  The rules for endpoints should as uniform as possible and presented in one general "API" page.  Of course, we should also have a link from the GSQL endpoints page pointing to the general point.

I recommend that you
1)  review the rules on this page to see if they are correct.
2) add a sentence on the GSQL endpoints page advising them where to look for authentication information.

I have one question:
It is not clear in the examples why you sometimes use username/password and sometimes use a token? It is random, or is there a pattern?

I personally do not like to show -u tigergraph:tigergraph in every example. It is not really a good practice.  I would have a note in the Authentication section telling them that this is the default username/password. It should ONLY be user for a private personal deployment where you are not concerned about security.  We tell users that they really should change the password, for security reasons.

dadongwang-tg:
1. let's change to use <jwt-token> in all examples @ArthurXie-tg 
2. for GSQL API there is 3 authentication methods, while for RESTPP API, there is only one authentication method, let's state this as well

PR #749, opened 2025-03-18
DOC-2605 Doc for cluster/node IP/Hostname change[3.10]

srinadhmamidalaTG:


modules/installation/pages/change-port
```diff
@@ -0,0 +1,37 @@
...
+
+[source,bash]
+----
+gadmin init etcd
+gadmin stop all
+----
+
+3. Make your IP changes
+
+4. Edit the .tg.cfg file, replacing the old IP/hostname information with the new information.
```
victorleeTG, line 21: Search for the JSON field called "Hostlist".

victorleeTG (changes requested):

PR #753, opened 2025-03-19
DOC-2449-Header `PROFILE` added

Tushar-TG-14:


modules/API/pages/built-in-endpoints.adoc
```diff
@@ -2745,6 +2745,16 @@ The Run Query endpoint accepts a number of headers that allow you to configure t
 |===
 |Name |Description
 
+|`PROFILE`
+|This header is used for query profiling. Currently, it supports only the value `BASIC`.
+When activated, query profiling allows users to view and analyze detailed information about query execution.
+By default, the value is empty, and profiling is not activated.
```
victorleeTG:
```suggestion
|Specifies the type of query profiling. Currently, only `BASIC` is supported.
Query profiling profiles detailed information about query execution for users to view and analyze.
Default: empty (no profiling).
```
You do not need or want a complete sentence at the beginning of a Description in a table like this. The table structure already implies: "This <item> is <description>."  See the existing rows in the table.

Headers are a standard concept for REST APIs, so you don't need any example here. Notice that none of the other rows have an example.

BUT, we need to tell users where to look to see the query profile output.  I think the engineer posted an example, but I haven't looked at it yet. I assume you are adding a section somewhere else to describe the query profiling feature. After you do that, add an xref here to link to that section.

victorleeTG (changes requested):

modules/API/pages/built-in-endpoints.adoc
Tushar-TG-14: @victorleeTG Noted.

PR #773, opened 2025-03-30
DOC-2359-incremental-backup

victorleeTG:


modules/backup-and-restore/pages/incremental-backup.adoc
```diff
@@ -0,0 +1,105 @@
...
+Suppose you made the following backups, at the times indicated by these tags:
+
+[console]
+----
+full-weekly-jan-01
+incr-daily-jan-02
+incr-daily-jan-03
+incr-daily-jan-04
+----
+If you want to restore to January 3, then you need to run the following sequence of restore operations:
```
runlilong, line 90: we don't need to run 3 restore commands. 

Simply run `gadmin backup restore incr-daily-jan-03`; the command itself will figure out its dependency backups and properly restore them (restore full backup first, and then restore all incremental backups in one time, as incremental backups are kafka message, we can write them back together).

PR #782, opened 2025-04-01
DOC-2460 Rolling Upgrade for maintenance releases

Tushar-TG-14:


modules/installation/pages/upgrade.adoc
```diff
@@ -85,6 +85,128 @@ $ gadmin restart all -y
...
 . Enable CRR on the DR cluster.
 
+[#_rolling_upgrade]
+== Rolling Upgrade (Preview)
+
+A rolling upgrade allows you to upgrade your TigerGraph installation with minimal downtime by upgrading nodes in batches, ensuring that at least one full replica of the cluster is online throughout the process. 
+
+[NOTE]
+====
+Rolling Upgrade is only available for HA clusters (High Availability) and is supported for maintenance releases (e.g., 4.2.0 to 4.2.1, 4.3.1 to 4.3.3).
```
```suggestion
Rolling Upgrade is only available for HA clusters (High Availability) and when making a maintenance update (e.g., 4.2.0 to 4.2.1, 4.3.1 to 4.3.3). 
```

modules/installation/pages/upgrade.adoc
```diff
@@ -85,6 +85,128 @@ $ gadmin restart all -y
...
+Run the following installation script with the upgrade flag (`+-Ur+`) that was extracted from the tar file:
+
+[source, gsql]
+----
+$ ./install.sh -Ur
+----
+
+. Switching to New Version
+
+After the binaries and configuration files are installed, user will be prompted with the following message:
```
```suggestion
After the binaries and configuration files are installed, the following prompt message will appear on the console:
```

modules/installation/pages/upgrade.adoc
```diff
@@ -85,6 +85,128 @@ $ gadmin restart all -y
...
+----
+To switch platform to new version, please run 'gadmin upgrade 4.2.1'
+----
+
+. y/Y Option
+
+If users selects `y/Y' the installer will automatically switch TigerGraph to the latest maintenance version. 
+If any errors occur, the upgrade process will terminate immediately. Follow the instructions in the prompts to manually continue the upgrade or revert to the original version.
+
+. Option `-n`
```
victorleeTG: I think this option is for the original command, e.g. ` ./install.sh -Ur -n`
Is that right? It's not clear from the document.

It is confusing to have it in this location, because it is seems be a 3rd choice from the y/N option.
Maybe use a heading instead of a numbered list.

modules/installation/pages/upgrade.adoc
```diff
@@ -85,6 +85,128 @@ $ gadmin restart all -y
...
+*Manual Upgrade*: Execute the following commands to restart services in sequence:
+
+[source.wrap,bash]
+----
+gadmin restart GUI#1 ADMIN#1 NGINX#1 EXE_1 IFM#1 KAFKASTRM-LL_1 RESTPP#1 DICT#1 KAFKACONN#1 GSE_1#1 GPE_1#1 CTRL#1 DICT#2 KAFKACONN#2 RESTPP#2 IFM#2 GUI#2 EXE_2 ADMIN#2 KAFKASTRM-LL_2 CTRL#2 GSE_1#2 NGINX#2 GPE_1#2 NGINX#3 GPE_1#3 GUI#3 KAFKASTRM-LL_3 DICT#3 KAFKACONN#3 RESTPP#3 EXE_3 ADMIN#3 CTRL#3 GSE_1#3 IFM#3 --no-dep -y
+----
+
+*Long-Running Operations*
+
+Before starting the upgrade process, ensure that long-running operations are allowed by running the following command:
```
victorleeTG: If this step is for "before starting the upgrade", shouldn't this be at the beginning, as a Prerequisite?

modules/installation/pages/upgrade.adoc
```diff
@@ -85,6 +85,128 @@ $ gadmin restart all -y
...
+$ gadmin upgrade --resume
+----
+
+. Revert to Previous Version
+
+To revert to your original version, follow these steps:
+
+[source.wrap,bash]
+----
+$ gadmin config set System.AppRoot /home/tigergraph/tigergraph/app/4.2.0
```
```suggestion
$ gadmin config set System.AppRoot /home/tigergraph/tigergraph/app/<old version>
```

modules/installation/pages/upgrade.adoc
```diff
@@ -85,6 +85,128 @@ $ gadmin restart all -y
...
+$ gadmin config apply -y
+$ gadmin restart all -y
+----
+
```
```suggestion
where <old version> is the version you were running before the upgrade, e.g. `4.2.0`.
```

modules/installation/pages/upgrade.adoc
```diff
@@ -85,6 +85,128 @@ $ gadmin restart all -y
...
+
+Welcome to the TigerGraph platform installer!
+
+[PROGRESS]: 08:10:44 Upgrade TigerGraph platform from existing one ... 
+[NOTE    ]: Obtained version of exist platform: 4.2.0 [OK] 
+[NOTE    ]: Obtained version of new platform: 4.2.1 [OK] 
+----
+
+. Handling Downtime
```
victorleeTG: I really think numbered list (the "dot" syntax) is not the right choice. We should use section headings instead. This is not a strict sequence of steps. There is a "soft" sequence of 
1. Preparation
2. Running rolling upgrade
3. Reverting if necessary/desired
but I still wouldn't number them.

Within "Running rolling upgrade" there is a mini sequence, because it's a 2-phrase process. We should actually describe this 2-phase process in the intro, so readers understand what is happening and understand their choices for how to run `./install.sh -Ur`:
1. Install the new version on one cluster node at a time but continue to run the old version.
2. Switch all nodes from the old version to the new version.

victorleeTG (changes requested):

modules/installation/pages/upgrade.adoc
```diff
@@ -85,6 +85,136 @@ $ gadmin restart all -y
...
+Upgrading from 4.2.0 to 4.3.0 or any other major version is not supported.
+====
+
+During a rolling upgrade, several operations are not permitted to ensure a smooth upgrade process. 
+These operations include installing queries, running schema changes, dropping all data, using `gsql --reset`, creating graphs, performing backup/restore operations, clearing the graph store, exporting/importing graphs, installing functions, changing TigerGraph configurations, starting or stopping services, setting up CRR (though CRR will continue working if already set up), running `gadmin init` for cluster initialization, upgrading services with `gadmin upgrade`, and performing service resets with `gadmin reset`. 
+While the rolling upgrade is in progress, loading jobs and running queries will continue to function; however, queries requiring all replicas to be online may fail, in which case you can retry them in a round-robin manner. 
+Please note that occasional drops in Queries Per Second (QPS) may occur during the upgrade. Additionally, services like Zookeeper (ZK), Kafka, and ETCD will not be upgraded during this process.
```
victorleeTG: Let's not use the word "drop" if we only mean a "slowdown". 
"Drop" usually has a more serious meaning with databases.

modules/installation/pages/upgrade.adoc
```diff
@@ -85,6 +85,136 @@ $ gadmin restart all -y
...
+
+The rolling upgrade follows a two-phase process:
+
+. *Install the new version on one cluster node at a time while continuing to run the old version*. This ensures that your cluster remains online with at least one replica available at all times during the upgrade.
+
+. *Switch all nodes from the old version to the new version*, completing the upgrade across the entire cluster. You can choose to run the upgrade manually by selecting the `y/Y` option, or automatically by using the `-n` flag to bypass manual input.
+
+=== Rolling Upgrade Process
```
victorleeTG, line 110: We should also tell the users to look at the general Prerequisites we have (e.g. earlier on this page) for any backup.

modules/installation/pages/upgrade.adoc
```diff
@@ -85,6 +85,136 @@ $ gadmin restart all -y
...
+After the binaries and configuration files are installed, the following prompt message will appear on the console:
+
+[console]
+----
+"Do you want to switch platform to the new version now (it can be delayed to a later time): y/N"
+----
+
+==== n/N option
+
+Selecting n/N will guide the user through the manual steps to switch to version 4.2.1, assuming the user is upgrading from 4.2.0.
```
victorleeTG, line 146: Why not just say "the new version"?

PR #830, opened 2025-05-20
DOC-2792-Updated Rolling Upgrade Page

Tushar-TG-14:


modules/installation/pages/upgrade.adoc
```diff
@@ -101,31 +101,18 @@ These operations include installing queries, running schema changes, dropping al
...
 ==== Download and Extract Latest Version
 
 * Download the latest maintenance release of TigerGraph.
 * Extract the tar file on your system.
 
-==== Run Installation Script with Upgrade Flag
+==== Run Installation Script with Rolling Upgrade Flag
 
-Run the following installation script with the upgrade flag (`+-Ur+`) that was extracted from the tar file:
+To perform a rolling upgrade, run the installation script with the upgrade flag (`+-U+`) and the rolling upgrade option (`+-r+`).
```
victorleeTG, line 115: @Tushar-TG-14 I don't understand the purpose of the "+" symbols in the formatted text.
I suppose it is a formatting marker, but do we actually need it?

modules/installation/pages/upgrade.adoc
```diff
@@ -199,28 +186,67 @@ Welcome to the TigerGraph platform installer!
...
+$ gadmin config set System.AppRoot /home/tigergraph/tigergraph/app/<old version>
+$ gadmin config apply -y
+$ gadmin restart all -y
+----
+
+Replace `<old version>` with the version you were running before the upgrade (e.g., `4.2.0`).
+
+=== Rolling Upgrade Failure Without Error
+
+Sometimes the rolling upgrade may fail without showing any error, like if the controller node is restarted manually.
```
```suggestion
Sometimes the rolling upgrade may fail without showing any error, like if the controller node is restarted manually. 
```
"Fail without error" is a scary situation for a user. I would like to developers to give us some more information about when this could happen, what could be the type of failure, and how the user would eventually realize that it failed.
This is not a small wording problem. The developers should have more awareness of how a user would feel about such statements.

modules/installation/pages/upgrade.adoc
```diff
@@ -199,28 +186,67 @@ Welcome to the TigerGraph platform installer!
...
+$ gadmin config apply -y
+$ gadmin restart all -y
+----
+
+Replace `<old version>` with the version you were running before the upgrade (e.g., `4.2.0`).
+
+=== Rolling Upgrade Failure Without Error
+
+Sometimes the rolling upgrade may fail without showing any error, like if the controller node is restarted manually. 
+When this happens, you can either *manually switch to the new version* yourself or *revert to the previous version*.
```
victorleeTG: @Tushar-TG-14 If the upgrade failed, why would we switch to the new version? It failed.

victorleeTG (changes requested):

modules/installation/pages/upgrade.adoc
```diff
@@ -199,28 +186,75 @@ Welcome to the TigerGraph platform installer!
 [PROGRESS]: 08:20:33 Switching to new version 4.2.1 successfully...
 ----
 
-=== Managing Service Disruption
+=== Managing Service Disruption During Rolling Upgrade
 
-During the rolling upgrade, expect a temporary service disruption. If the upgrade fails or the connection is lost, user can *resume* the upgrade process by running:
+During a rolling upgrade, temporary service interruptions are expected. If the upgrade process fails or the connection is unexpectedly lost, you can resume the upgrade by running:
```
```suggestion
During a rolling upgrade, the TigerGraph service may be briefly disrupted, but it should resume on its on.
```
Comments: The developer didn't say it would resume on its own, but I'm guessing/hoping that is the case?
For very briefly disruptions, the TigerGraph system will often automatically "retry", so all of the requests that the user submitted still run; they just were delayed.  It would be helpful if the developer tells if if the user might need to resubmit some requests.  Also, I think it is better to tell the user to avoid long queries or loading jobs during an upgrade.  If you have to restart a long job, it's more of an inconvenience and lot of efficiency than restarting a short job.

modules/installation/pages/upgrade.adoc
```diff
@@ -199,28 +186,75 @@ Welcome to the TigerGraph platform installer!
 [PROGRESS]: 08:20:33 Switching to new version 4.2.1 successfully...
 ----
 
-=== Managing Service Disruption
+=== Managing Service Disruption During Rolling Upgrade
```
victorleeTG: I would make this a NOTE, not a section.

modules/installation/pages/upgrade.adoc
victorleeTG: As for what do to **if the upgrade fails**, that is handled in another section and is more complicated. It is confusing/inconsistent/wrong to give simple instructions here.

As for what to do if "connection is unexpectedly lost", is it not clear what we mean by "connection". "Connection lost" is more of a symptom than a root cause.  "TigerGraph service disruption" could be considered a "lost connection", so the situations are overlapping.  In general, I think we should take this out here. Just address it properly and fully in the "Recovering from an Incomplete Rolling Backup"

modules/installation/pages/upgrade.adoc
```diff
@@ -199,28 +186,75 @@ Welcome to the TigerGraph platform installer!
...
+During a rolling upgrade, temporary service interruptions are expected. If the upgrade process fails or the connection is unexpectedly lost, you can resume the upgrade by running:
 
 [source.wrap,bash]
 ----
 $ gadmin upgrade --resume
 ----
 
-If the upgrade encounters an issue, the error prompt will guide you on how to continue the upgrade manually. 
-Use the following command if you need to manually finish the upgrade:
+=== Rolling Upgrade Failure with Error
```
victorleeTG: Change the title to something like "Recovering from An Incomplete Rolling Backup".
The subsection should be on Determining that Your Rolling Backup Didn't Finish.
The next section(s) should be on your two options: Continue, but as a manual, offline backup. Or, revert back to the old version. If the disruption is not expected to happen again, they could them retry.

victorleeTG (changes requested):

modules/installation/pages/upgrade.adoc
```diff
@@ -199,37 +186,63 @@ Welcome to the TigerGraph platform installer!
 [PROGRESS]: 08:20:33 Switching to new version 4.2.1 successfully...
 ----
 
-=== Managing Service Disruption
+[NOTE]
+====
+During a rolling upgrade, the TigerGraph service may be briefly disrupted but should resume on its own.
+====
+
+=== Recovering from an Incomplete Rolling Backup
```
```suggestion
=== Recovering from an Incomplete Rolling Upgrade
```
Probably my typo

modules/installation/pages/upgrade.adoc
```diff
@@ -199,37 +186,63 @@ Welcome to the TigerGraph platform installer!
...
+gadmin upgrade --resume
 ----
 
-If the upgrade encounters an issue, the error prompt will guide you on how to continue the upgrade manually. 
-Use the following command if you need to manually finish the upgrade:
+If you receive a clear success or error message, follow the instructions accordingly.
+If you receive no message and the upgrade process appears stalled, treat this as a silent failure.
 
-Execute the following command to restart services in sequence:
+=== How to Proceed After an Incomplete Rolling Backup
```
victorleeTG: This headig should not be at the same heading level as the previous one. My intent for the previous section ("Recovering...") was to cover both diagnosing problems and handling problems. So it is the parent section, which could have subsections.

The workflow is this:
1. Determining the Status of Your Rolling Upgrade
Here you see if your rolling upgrade finished successfully or it not.  For example, if there was a network interruption during the rolling upgrade, the process may have stopped in an incomplete phase.

There are two types of incomplete rolling backups.

2. Handling an Incomplete Rolling Upgrade
(Here you list out the different situations and how to handle them.
I think "offline upgrade" is a more clear terminology that "manual".  From a user perspective, the level of automation is the same for rolling upgrade vs "manual upgrade": just run one command and wait. But offline upgrade requires that all other activity be stopped.)

victorleeTG (changes requested):

modules/installation/pages/upgrade.adoc
```diff
@@ -199,37 +186,67 @@ Welcome to the TigerGraph platform installer!
...
 $ gadmin restart all -y
 ----
 
-where <old version> is the version you were running before the upgrade, e.g. `4.2.0`.
+Replace `<old version>` with the version you were running before the upgrade (e.g., `4.2.0`).
+
+[NOTE]
+====
+* If the issue that caused the upgrade to stop is fixed and not expected to happen again, you can try running the rolling upgrade again as usual.
+* Be aware that both manual finishing and reverting to the previous version may cause temporary service downtime.
```
```suggestion
* If you choose to upgrade offline or revert back to the previous version, you should halt database operations until the version change is complete.
```

victorleeTG (approved):

PR #837, opened 2025-06-17
DOC-2842-bug fixes for 3.10.4

Tushar-TG-14:


modules/release-notes/pages/index.adoc
```diff
@@ -169,6 +169,24 @@ See xref:3.10@tigergraph-server:user-access:rbac-row-policy/row-policy-privilege
 
 == Fixed issues
 
+=== Fixed and Improved in 3.10.4
+
+==== Functionality
+
+* Fixed an issue where the `/codecheck` endpoint returned an empty response for queries with syntax or semantic errors, which previously allowed incorrect queries to be installed (GLE-10432).
```
victorleeTG: This one confused me: it sounds like 2 separate problems (empty response, bad queries being installed).
I finally figured out that /codecheck is an *internal endpoint* and that we use it to decide which queries to install. 
That detail was not clear.

Please improve by either (1) giving LESS detail: don't mention /codecheck.  Describe only what users see, or (2) giving MORE detail. Explain that /codecheck is internal and how it is used.

modules/release-notes/pages/index.adoc
```diff
@@ -169,6 +169,24 @@ See xref:3.10@tigergraph-server:user-access:rbac-row-policy/row-policy-privilege
...
+* Fixed a backward compatibility issue that caused system unavailability or schema check failures after upgrading when the keyword `function` was used in vertex attributes (GLE-9840).
+* Fixed an *Out-of-Memory (OOM)* issue that occurred during file loading operations (TP-8151).
+* Fixed an issue where vertex attributes loaded with invalid values when a loading job inserted vertices using reducer operators such as `and()`, `ignore_if_exists()`, `max()`, or `min()` (CORE-4866).
+
+==== Improvements
+
+* Upgraded Nginx to version 1.26 to enhance compliance and provide an updated underlying service component (TP-8129).
```
victorleeTG: In general, you don't need to explain why we upgrade components. Only if there was a specific new feature (in Nginx) we used or a specific bug (in Nginx) that was fixed.

Otherwise, just say "Upgraded Nginx to version 1.26."

modules/release-notes/pages/index.adoc
```diff
@@ -169,6 +169,24 @@ See xref:3.10@tigergraph-server:user-access:rbac-row-policy/row-policy-privilege
...
+
+==== Security
+
+* Removed GraphStudio exploration results and loading job sample data from `gui/data.json` to prevent potential leakage of Personally Identifiable Information (PII) during schema export (APPS-3946).
+* Fixed PII exposure in GSE INFO logs by masking vertex primary IDs to enhance data privacy (CORE-4860).
```
victorleeTG: These are similar changes. I like the first one but not the second one.
The first one says what we did, to prevent a "potential" problem.
Please rewrite the second one to follow the same form.

Security and PII sensitive problems. How would you feel if your bank said we accidentally published your bank account information to users who don't have the proper privilege?

victorleeTG (changes requested):

modules/release-notes/pages/index.adoc
```diff
@@ -169,6 +169,33 @@ See xref:3.10@tigergraph-server:user-access:rbac-row-policy/row-policy-privilege
...
 == Fixed issues
 
+=== Fixed and Improved in 3.10.4
+
+==== Functionality
+
+* Fixed an issue where incorrect queries with syntax or semantic errors were allowed to be installed (GLE-10432).
+* Fixed a backward compatibility issue that caused system unavailability or schema check failures after upgrading when the keyword `function` was used in vertex attributes (GLE-9840).
+* Fixed an *Out-of-Memory (OOM)* issue that occurred during file loading operations (TP-8151).
+* Fixed an issue where vertex attributes loaded with invalid values when a loading job inserted vertices using reducer operators such as `and()`, `ignore_if_exists()`, `max()`, or `min()` (CORE-4866).
```
```suggestion
* Fixed an issue where a loading job inserted invalid vertex attribute values when reducer operators such as `and()`, `ignore_if_exists()`, `max()`, or `min() were used` (CORE-4866).
```
The wording "vertex attributes loaded with invalid values" confused me. Did we mean:
* vertex attributes that loaded with invalid values - is the group of things we are talking.  This was how I interpreted it first, but then the rest of the sentence didn't make sentence grammatically.
* vertex attributes got loaded with invalid values - and that is the problem.  This is what you meant.
So I suggested a different wording.

modules/release-notes/pages/index.adoc
```diff
@@ -169,6 +169,33 @@ See xref:3.10@tigergraph-server:user-access:rbac-row-policy/row-policy-privilege
...
+* Fixed data rejection during loading jobs by enhancing timestamp conversion to support a wider range of dates, including those prior to 1970-01-01 (GLE-10579).
```
victorleeTG: This is not really what we did. We did not change the behavior of `gsql_ts_to_epoch_seconds()`; it continues to only work for dates from 1970, because it returns only positive numbers.
Instead, there is a feature change to introduce some new features (so the customer can use those new functions instead, if they want to.)

victorleeTG: So, GLE-10579 should not be in the Issues Fixed list at all. The new functions should be listed as New Features.

modules/release-notes/pages/index.adoc
```diff
@@ -169,6 +169,33 @@ See xref:3.10@tigergraph-server:user-access:rbac-row-policy/row-policy-privilege
...
+* Fixed an issue where retrieving the query list in Write Queries would get stuck when creating a large query (GLE-10461).
```
```suggestion
* Fixed a GraphStudio issue where retrieving the query list in Write Queries would get stuck when creating a large query (GLE-10461).
```
I was confused by the wording because there is no `Write Queries ` command in GSQL. I looked at the ticket and saw that it was for GraphStudio, where Write Queries is the name of one of the work panels.

modules/release-notes/pages/index.adoc
```diff
@@ -169,6 +169,33 @@ See xref:3.10@tigergraph-server:user-access:rbac-row-policy/row-policy-privilege
...
+* Fixed an issue where users could not locate a vertex in the Graph Studio result when the executed query only printed edges (APPS-3991).
```
```suggestion
* Fixed an issue where users could not locate a vertex in the GraphStudio result when the executed query only printed edges (APPS-3991).
```
Our job is to fix it when other's use the wrong product name.

modules/release-notes/pages/index.adoc
```diff
@@ -169,6 +169,33 @@ See xref:3.10@tigergraph-server:user-access:rbac-row-policy/row-policy-privilege
...
+* Fixed an issue where an incorrect 'query not running' error message appeared when aborting a query in the Admin Portal on multi-node systems (APPS-3962).
+* Fixed an issue where the delete query button was incorrectly enabled for queries that were only saved but not yet installed, causing an error when attempting to delete them (APPS-3917).
```
victorleeTG: Ask yourself: What delete button? What tool are the talking about? Please read the ticket and then specify what tool we are talking about.

modules/release-notes/pages/index.adoc
```diff
@@ -169,6 +169,33 @@ See xref:3.10@tigergraph-server:user-access:rbac-row-policy/row-policy-privilege
...
+* Fixed an issue where query changes were lost or reverted when clicking the save button immediately after editing or fixing syntax errors (APPS-3916).
```
victorleeTG: Same: what tool's Save button?

modules/release-notes/pages/index.adoc
```diff
@@ -169,6 +169,33 @@ See xref:3.10@tigergraph-server:user-access:rbac-row-policy/row-policy-privilege
...
+* Fixed issue where codecheck failed to prevent the installation of a query with syntax errors in Write Queries, by catching unexpected exceptions and reporting errors, rather than skipping the query (GLE-10467).
```
```suggestion
* Fixed issue where queries with certain syntax errors disappeared during installation, because the internal codecheck utility returned an empty result instead of an error message (GLE-10467).
```
Initially, I was going to suggest that we shorten it to "Fixed issue where codecheck failed to prevent the installation of a query with syntax errors in Write Queries" because internal details about reporting errors and skipping a query sound like the obvious things that checking query syntax is supposed to do.

Then I remembered that the user problem is Disappearing Queries.  I thought we should mention this, and THEN there is way to connect the dots.  It the error checking produced an empty output, for "certain" errors (not all errors), then that explains how a query could disappear.

modules/release-notes/pages/index.adoc
```diff
@@ -169,6 +169,33 @@ See xref:3.10@tigergraph-server:user-access:rbac-row-policy/row-policy-privilege
...
+* Fixed issue with exporting only vertexset attributes to CSV when using the `print <vertexset> TO_CSV <file>;` statement (GLE-10521).
```
victorleeTG: This is not a Fixed Issue; this is a new feature/functionality.
While I'm still a little confused by engineering's explanation, I think this is the situation.
If A is a vertex set, then
* `PRINT A;`   # prints all attributes and accumulators of A. This was already supported but not documentation so we can still treat it like a new feature.
* `PRINT A TO_CSV myFile;` # NEW.  This case includes TO_CSV whereas the other cases do not. However, it will print only the attributes of A, not the accumulators. (The lack of accumulators is actually a shortcoming which I think the engineers did not realize until I asked them to specify the output details.)
* `PRINT A.*;`  # Not supported. NEW: produce an improved error message which suggests using PRINT A
* `PRINT A[A.*];`  # Not supported. NEW: produce an improved error message which suggests using PRINT A

modules/release-notes/pages/index.adoc
```diff
@@ -169,6 +169,33 @@ See xref:3.10@tigergraph-server:user-access:rbac-row-policy/row-policy-privilege
...
+
+==== Improvements
+
+* Upgraded Nginx to version 1.26 (TP-8129).
+* Removed authentication requirements for replaying requests in Disaster Recovery (DR) setups, streamlining Cross-Region Replication (CRR) operations (GLE-10601).
```
victorleeTG: As a user, I still don't know enough to understand how this affects me.
Is there now a step a can skip? What step?
Is this during regular CRR behavior (maintaining the remote replica) or during recovery (the primary cluster has failed)?

victorleeTG (changes requested):

victorleeTG (approved):

PR #854, opened 2025-08-07
DOC-2884 fixed the changes on rolling upgrade feature

Tushar-TG-14:


modules/installation/pages/upgrade.adoc
```diff
@@ -92,42 +92,74 @@ $ gadmin restart all -y
...
 
-A rolling upgrade allows you to upgrade your TigerGraph installation with minimal downtime by upgrading nodes in batches, ensuring that at least one full replica of the cluster is online throughout the process. 
+A rolling upgrade allows you to upgrade your TigerGraph installation with minimal downtime by upgrading nodes in batches, ensuring that at least one full replica of the cluster is online throughout the process.
 
-[NOTE]
-====
-Rolling Upgrade is only available for HA clusters (High Availability) and when making a maintenance update (e.g., 4.2.0 to 4.2.1, 4.3.1 to 4.3.3). 
-Upgrading from 4.2.0 to 4.3.0 or any other major version is not supported.
-====
+*Before You Begin*
```
victorleeTG: Why I don't like this title: Because it doesn't fit the style and purpose of this document. This document is part of a **reference manual** which primarily describes WHAT a feature is and secondarily how to use it, vs a **user guide** which primary describes HOW to perform a particular task. User guides are fine, but that's not what this is.
"Before you begin": who is "you" and before you begin what?  In a reference manual this doesn't make sense because we are focusing on what the feature IS (existential, always true), not on a reading performing some particular task.

modules/installation/pages/upgrade.adoc
```diff
@@ -92,42 +92,74 @@ $ gadmin restart all -y
...
+
+Rolling Upgrade is only available for HA (High Availability) clusters.
+It is supported only for maintenance updates (for example, 4.2.0 to 4.2.1, or 4.3.1 to 4.3.3). Upgrading from 4.2.0 to 4.3.0 or another major version is not supported.
```
```suggestion
It is supported only for maintenance updates (for example, 4.2.0 to 4.2.1, or 4.3.1 to 4.3.3). Upgrading from 4.2.x to 4.3.0 or another major version is not supported.
```

modules/installation/pages/upgrade.adoc
```diff
@@ -92,42 +92,74 @@ $ gadmin restart all -y
...
+
+During a rolling upgrade:
+
+* Loading jobs and running queries will continue to function.
+* Occasional slowdowns in queries per second (QPS) may occur.
+* TigerGraph does not upgrade Zookeeper (ZK), Kafka, or ETCD as part of this upgrade.
```
```suggestion
* Third-party components such as Kafka are not upgraded unless the Release Notes for the maintenance upgrade explicitly mention an upgrade.
```
This item is an "existential" item and not a "during" item. It belongs in the previous list,  not this list.

victorleeTG (changes requested):
Partial comments, review not complete

modules/installation/pages/upgrade.adoc
```diff
@@ -92,42 +92,74 @@ $ gadmin restart all -y
...
-* Export or import graph
-* Install function
+* `INSTALL QUERY`
+* `SCHEMA CHANGE`
+* `DROP ALL`
+* `gsql --reset`
+* `CREATE GRAPH`
+* `CLEAR GRAPH STORE`
+* `EXPORT` or `IMPORT GRAPH`
+* `INSTALL FUNCTION`
```
victorleeTG: The original list was written as English descriptions of features; that is why it was in lowercase.
If we capitalize the items, then we are writing code, and the code has to be correct. Now if I look at it from a code perspectively, I see a lot of contextual inconsistencies. That is, the developer created a problem by holding us to a higher standard without fixing the code.

modules/installation/pages/upgrade.adoc
```diff
@@ -92,42 +92,74 @@ $ gadmin restart all -y
...
-Upgrading from 4.2.0 to 4.3.0 or any other major version is not supported.
-====
+*Before You Begin*
+
+Rolling Upgrade is only available for HA (High Availability) clusters.
+It is supported only for maintenance updates (for example, 4.2.0 to 4.2.1, or 4.3.1 to 4.3.3). Upgrading from 4.2.0 to 4.3.0 or another major version is not supported.
+
+During a rolling upgrade:
+
+* Loading jobs and running queries will continue to function.
```
```suggestion
* Loading jobs and running queries will continue to function, with some limitations described in the [next section].
```

PR #868, opened 2025-09-25
DOC-2923: Add documentation for `GET /gsql/v1/schema/version` API

Tushar-TG-14:


modules/API/pages/gsql-endpoints.adoc
```diff
@@ -1769,13 +1769,61 @@ Sample Response::
...
+==== Example:
+
+[.tabs]
+====
+Sample Request::
++
+--
+[source,bash]
+----
+curl --user tigergraph:tigergraph http://<tigergraph-host>:14240/gsql/v1/schema/version
```
victorleeTG: The format of curl command example is not consistent with the style of the existing endpoint examples.

modules/API/pages/gsql-endpoints.adoc
```diff
@@ -1769,13 +1769,61 @@ Sample Response::
...
+
+Sample Response::
++
+--
+[source.wrap,console]
+----
+{
+  "error": false,
+  "message": "",
+  "version": 1
```
victorleeTG, line 1808: Is the version always an integer? We should state what the data type is.

victorleeTG (changes requested):

PR #869, opened 2025-09-26
DOC-3029 Add `GSQL.AllowInterpretedQueries` configuration parameter

Tushar-TG-14:


modules/reference/pages/configuration-parameters.adoc
```diff
@@ -542,6 +542,17 @@ online |`300`
 |GSQL.QueryPlanCache.Enable |Enables the xref:{page-component-version}@gsql-ref:querying:query-optimizer/query-plan-cache.adoc[query plan cache], default is `true` |`true`
 
 |GSQL.QueryPlanCache.Capacity |Maximum number of xref:{page-component-version}@gsql-ref:querying:query-optimizer/query-plan-cache.adoc[queries in the cache]. Range from 1 : 100,000. default is 10,000. |`10000`
+
+|GSQL.AllowInterpretedQueries 
+a|Enables or disables all interpreted (ad hoc) GSQL queries for the cluster. Default: `true`.
```
xun-gao-tigergraph: Is there a redundant "a"?

xun-gao-tigergraph: I think there should not be `cluster` since the environment may be a single node environment.

modules/reference/pages/configuration-parameters.adoc
```diff
@@ -633,6 +644,7 @@ request timeout in milliseconds |`600000`
 
 |Gadmin.StopServiceDefaultTimeoutMS |The stop one service default
 request timeout in milliseconds. Default 200000 (200s).|`30000`
+
```
xun-gao-tigergraph, line 646: redundant space?

modules/reference/pages/configuration-parameters.adoc
Tushar-TG-14: @xun-gao-tigergraph Thanks for the review!

The a| is intentional here, it’s required in AsciiDoc to allow rich content (like bullet lists) inside a table cell. You can check the “Display the rich diff” view to see how it renders correctly.
Good point on “cluster”. I’ve updated it to “TigerGraph environment” to also cover single-node setups.
I’ve also cleaned up spacing for consistency.

Please take another look when you have a moment.

PR #872, opened 2025-09-30
DOC-2921 support sampling local file [4.3]

AyeshaFirdausTG:


modules/API/pages/gsql-endpoints.adoc
```diff
@@ -2915,6 +2915,61 @@ Sample Response::
 ----
 --
 ====
+=== Get the sample data of local files
+`POST /gsql/v1/sample-data`
+
+Get the sample data of local files
```
victorleeTG, line 2995:
```suggestion
Get the first N of lines of data from a local file, as a sample
```
"Sample" is a generic, non-specific concept.  To a statistician, "sample" actually suggests that the data is chosen randomly.  But I *assume* our selection is not random. We are selecting the first N lines. Is that correct?
We should provide a clearly explanation of how the data is chosen.

modules/API/pages/gsql-endpoints.adoc
```diff
@@ -2915,6 +2915,61 @@ Sample Response::
 ----
 --
 ====
+=== Get the sample data of local files
```
victorleeTG, line 2992:
```suggestion
=== Get sample data from a local file
```
"The" is incorrect, because there is not just one possible or correct sample of data.
The data is coming from only one file, not "files"

victorleeTG (changes requested):

PR #880, opened 2025-10-29
DOC-2777 CRR observability

AyeshaFirdausTG:


modules/cluster-and-ha-management/pages/crr-monitoring.adoc
```diff
@@ -1 +1,115 @@
...
+| Metric | `tigergraph_crr_replayed_timestamp` | topic | Timestamp of replayed record in DR
+|===
+
+[NOTE]
+====
+TigerGraph provides raw metrics for CRR topic replication and replay, but does not include built-in alerting capabilities.
+
+Replication and replay lag may vary depending on cluster configuration, network bandwidth, and latency. Users are advised to define alerting rules in their own monitoring systems (for example, Prometheus).
+
+A CRR topic can be considered abnormal if any of the following conditions occur:
```
pingxieTG: This is an example, not a rule that must be followed.

modules/cluster-and-ha-management/pages/fail-over.adoc
```diff
@@ -1,44 +1,150 @@
...
+
+==== 4. Forced Promotion for Emergency Scenarios
+In urgent cases, administrators can use the `--force` flag along with `--promote` to bypass synchronization checks.
+
+[source,bash]
+----
+gadmin crr stop --promote --force -y
+----
+
+Use this option only during emergencies. Forced promotion may lead to data loss and will require a full backup and restore to re-establish the DR cluster.
```
pingxieTG, line 49: We can further explain what “emergency” means — for example, when the Primary Cluster is down and we need to immediately switch the DR to Primary to ensure service continuity.

modules/cluster-and-ha-management/pages/fail-over.adoc
```diff
@@ -1,44 +1,150 @@
...
+gadmin crr stop --promote --dump-checkpoint <path>
 ----
 
-== Set up a new DR cluster after failover
+Validation outcomes:
+
+* **Validation passes:**
+Clusters were fully synchronized. A fast-switch occurs, converting the cluster to a DR instance instantly without data transfer.
+* **Validation fails:**
+The failover occurred during replication lag (e.g., when `--force` was used). A fast-switch is not possible, and a full backup and restore is required.
```
pingxieTG: e.g., when `--force` was used, or when new data is inserted into the old Primary Cluster after the failover

modules/cluster-and-ha-management/pages/crr-monitoring.adoc
```diff
@@ -1 +1,116 @@
...
+
+[source,console]
+----
+gadmin config entry System.Metrics.CRRIntervalSec
+----
+
+[#cli_status_query]
+== CLI-based CRR status query
+
+TigerGraph introduces a new `gadmin` sub-command to display detailed CRR metrics.
```
victorleeTG, line 20:
```suggestion
Users can run a `gadmin` command to get detailed CRR metrics.
```
You should not mention TigerGraph (the company). You should not say "new", because that will be outdated when 4.3 is no longer the newest version.
You should simply describe the command, e.g.
"Users can run a `gadmin` command to get detailed CRR metrics."

modules/cluster-and-ha-management/pages/crr-monitoring.adoc
```diff
@@ -1 +1,116 @@
...
+| `-v`, `--verbose` | Displays detailed metrics, including source and target offsets and timestamps.
+|===
+
+
+[#topic_metrics]
+=== CRR Topic Metrics 
+The following table provides the definitions of topic metrics.
+
+|===
+| Field Name | Description | Notes | Verbose
```
victorleeTG, line 50: The interpretation of the "verbose" column is not clear.
What does verbose = false mean for a field?  (I think "false" is not what you really mean.)
What does verbose = true mean for a field?

A more logical approach would be to call the column "shown when".
Then, the values are either
* always
* --verbose
* 
The nice thing about this approach is that it 
* very clear
* easily supports multiple levels of verbocity or more complex conditions. Just describe the condition.

modules/cluster-and-ha-management/pages/fail-over.adoc
```diff
@@ -1,44 +1,152 @@
 = Fail over to the DR cluster
 //:page-aliases: crr:fail-over.adoc
 
+
 In the event of catastrophic failure that has impacted the full cluster due to Data Center or Region failure, the user can initiate the failover to the Disaster Recovery (DR) cluster.
-This is a manual process.
 
-Run the following commands to make configuration changes on the DR cluster to upgrade it to the primary cluster.
+Starting from TigerGraph 4.3.0, failover to a Disaster Recovery (DR) cluster is simplified into a single command.
```
victorleeTG, line 7:
```suggestion
Starting from TigerGraph 4.3, failover to a Disaster Recovery (DR) cluster is simplified into a single command.
```

modules/cluster-and-ha-management/pages/fail-over.adoc
```diff
@@ -1,44 +1,152 @@
...
+
+[[set-up-new-dr-cluster]]
 
-After you fail over to your DR cluster, your DR cluster is now the primary cluster. You may want to set up a new DR cluster to still be able to recover your services in the event of another disaster.
+== Set Up a New DR Cluster After Failover
+:description: Procedure for re-establishing high availability after DR promotion using fast-switch or backup restore.
 
-To set up a new DR cluster over the upgraded primary cluster:
+=== Overview
+TigerGraph v4.3 introduces a new **fast-switch** capability that allows administrators to quickly convert the former Primary cluster into a new DR cluster.
```
victorleeTG, line 68: Be consistent.  Earlier in the page you said, "version 4.3" and "TigerGraph 4.3" (no "v" before "4.3").

PR #885, opened 2025-11-07
DOC-2912: Added and created a new page "SSO with Kerberos"

Tushar-TG-14:


dadongwang-tg (approved):
overall looks good to me, @xun-gao-tigergraph please double review.

xun-gao-tigergraph (approved):

modules/user-access/pages/sso-with-kerberos.adoc
```diff
@@ -1 +1,223 @@
...
+
+You can match principals in two ways:
+
+* *Exact match*
++
+`principal=alice/admin@example.com`
+
+* *Regex match*
++
+`principal=.*/admin@example.com`
```
xun-gao-tigergraph: this should be `regex: principal=.*/admin@example.com`

xun-gao-tigergraph (changes requested):

modules/user-access/pages/sso-with-kerberos.adoc
```diff
@@ -1 +1,223 @@
...
+
+|`Security.Kerberos.Krb5Config`
+|Specifies the content or path to the `krb5.conf` file. Use `@/path/to/krb5.conf` to read from a file.
+|Empty
+
+|`Security.Kerberos.Principal`
+|The service principal name used by TigerGraph. Example: `HTTP/tg-host.example.com@REALM`
+|Empty
+
+|`Security.SSO.EnableDefaultRole`
```
xun-gao-tigergraph, line 65: This one is not only used by Kerberos, but also used by other SSO approaches, such as SAML, OIDC, LDAP, etc.

modules/user-access/pages/sso-with-kerberos.adoc
```diff
@@ -1 +1,223 @@
...
+== Troubleshooting
+
+|===
+|Issue | Description
+
+|Authentication fails
+|Check if your Kerberos ticket has expired (`klist`). Renew it with `kinit`.
+
+|User logs in with no privileges
+|User’s principal does not match any proxy rule and `EnableDefaultRole=false`. Add a matching proxy group or set the default role to true.
```
dadongwang-tg, line 203: here should be `EnableDefaultRole=true`? @xun-gao-tigergraph

xun-gao-tigergraph: No, it should be false here.

PR #896, opened 2026-01-19
Update index.adoc

AyeshaFirdausTG:


modules/security/pages/index.adoc
```diff
@@ -36,6 +36,101 @@ TigerGraph Server meets the following security compliance standards as certified
...
+This feature introduces SSL/TLS encryption for TigerGraph’s internal Kafka. It addresses the lack of native encryption and authentication for data in transit within the TigerGraph cluster.
+
+=== Basic Setup Workflow
+
+[source,bash]
+----
+gadmin config set Kafka.Security.SSL.Certificate @/path/to/cert.pem
+----
+
+* Set certificate-chain in PEM format.
```
pingxieTG, line 50: The description can be placed before the code block. Same for the following.
Btw, is the numbered list better than bullet points here?

modules/security/pages/index.adoc
```diff
@@ -36,6 +36,101 @@ TigerGraph Server meets the following security compliance standards as certified
...
+----
+gadmin config apply -y
+----
+
+[source,bash]
+----
+gadmin restart -y kafka
+----
+
+Kafka restarts with the new configurations.
```
pingxieTG, line 81: Sorry, here we should `gadmin restart all -y`: Restart Kafka and the corresponding client services.

PR #906, opened 2026-02-17
DOC-2996-locking-critical-operations-4.3

AyeshaFirdausTG:


modules/backup-and-restore/pages/incremental-backup.adoc
```diff
@@ -5,10 +5,12 @@ An *incremental backup* is designed to save storage space and to run more quickl
 TigerGraph's incremental backup feature also has an option to let you specify a backup tag other the default base.
 
 *Incremental backups* are related to but different than *differential backups*.
-A different backup saves since the most recent backup or restore event.
+A differential backup saves since the most recent backup or restore event.
```
victorleeTG:
```suggestion
A differential backup saves all the changes since the most recent *full backup* or restore event.
```

victorleeTG (changes requested):
@AyeshaFirdausTG The locking comments looking fine.

There is a wording mistake about differential backup.  It looks like it was wrong before, and your change didn't really fix the problem.

PR #939, opened 2026-06-02
DOC-3213: known issue for rolling upgrade [4.2.3]

Tushar-TG-14:


modules/release-notes/pages/index.adoc
```diff
@@ -351,6 +351,11 @@ Third-party Vulnerability impacting TigerGraph: None
 ¦ Description ¦ Found In ¦ Workaround ¦ Fixed In
 
 
+¦Rolling upgrade from 4.2.2 to 4.2.3 may fail when `Nginx.SSL.Enable=true`. The upgrade fails during the GSQL unblock step due to an internal HTTPS request failure.
```
qianzhang-tg: 1. This issue can occur when rolling upgrade to 4.2.3 is performed from a version between 4.2.0 and 4.2.2. Not just 4.2.2. Please reflect that in here.
2. Upgrade fails at the step of blocking GSQL long-running operations. Not at the unblock step.

PR #945, opened 2026-06-11
DOC-2733: Document UDF changes and query reactivation post-upgrade [4.2]

Tushar-TG-14:
Added important note about UDF changes after upgrades.

### DIFF
--- a/modules/installation/pages/upgrade.adoc
+++ b/modules/installation/pages/upgrade.adoc
@@ -572,14 +572,13 @@ It is recommended that users consult with the entire policy and adjust the UDF f
 For more reference please see:
 https://docs.tigergraph.com/tigergraph-server/current/security/#_udf_file_scanning[UDF File Scanning].
 
-[IMPORTANT]
-====
-If an upgrade changes built-in UDFs or if you update custom UDFs, some queries may be disabled.
+== Post-Upgrade Checks
 
-After the upgrade, review any disabled queries and reinstall them as needed.
+After the upgrade completes, verify that your queries are installed and enabled.
+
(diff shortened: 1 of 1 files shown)


modules/installation/pages/upgrade.adoc
```diff
@@ -572,6 +572,15 @@ It is recommended that users consult with the entire policy and adjust the UDF f
 For more reference please see:
 https://docs.tigergraph.com/tigergraph-server/current/security/#_udf_file_scanning[UDF File Scanning].
 
+[IMPORTANT]
```
victorleeTG: This is not where I would put this info.
This page has a major problem: It lacks a workflow.  This new content should be in the "Post-Upgrade Checks" section, but there is no such section yet.  The top of the page should have a CONCISE workflow outline (see Schema Change for an example). The overview is not the place for how-to details. It is where you list the major steps so users mentally understand the high level process, and know what is ahead. Sometimes there are multiple paths/options, and they need guidance for which path to follow.

Something like (with hyperlinks, of course):
1. Preparing to Upgrade
Check for compatibility issues, make a backup, pause the system if needed.
2. Perform the Upgrade
Simple upgrades, Two-stage upgrades, CRR upgrades
3. Post-Upgrade Checks
Verify that queries and loading jobs are still installed. Perform a data validation check.

victorleeTG (changes requested):

victorleeTG (approved):
@Tushar-TG-14 
Acceptable if you also add a workflow outline at the top of the page, so readers know that his subsection exists.

PR #949, opened 2026-06-18
disable-saml

AyeshaFirdausTG:


xun-gao-tigergraph (commented):
Should we also mention that TigerGraph validates the X509 certificate of SAML-based SSO?

AyeshaFirdausTG:
> Should we also mention that TigerGraph validates the X509 certificate of SAML-based SSO?

Is this something users should know about or take into consideration?

xun-gao-tigergraph (approved):

PR #953, opened 2026-06-30
DOC-3252-RESTPP-local-gpe-4.3

AyeshaFirdausTG:


modules/system-management/pages/workload-management.adoc
```diff
@@ -578,3 +578,22 @@ If no GPE satisfies the CPU threshold condition, REST++ falls back to the defaul
...
+
+You can target a specific GPE by directing your requests to the RESTPP instance running on that GPE's node.
+
+.Example: Enable Local GPE routing
+[source, console]
+----
+$ gadmin config set RESTPP.QueryRouting.Mode 2
+$ gadmin config apply
+$ gadmin restart RESTPP -y
+----
```
Tushar-TG-14, line 604: After the example add one sentence:
After restarting RESTPP, all incoming queries handled by that RESTPP instance are forwarded to its local GPE.

modules/system-management/pages/workload-management.adoc
```diff
@@ -578,3 +578,22 @@ If no GPE satisfies the CPU threshold condition, REST++ falls back to the defaul
 [source, console]
 $ gadmin config entry RESTPP.QueryRouting.TargetSelectionCPUThreshold 40
 $ gadmin config entry RESTPP.QueryRouting.Mode 1
+
+
+
+=== Local GPE Routing
```
Tushar-TG-14, line 590: The page currently jumps
Round Robin
↓
CPU Load Aware
↓
Local GPE
without actually telling users how the three modes compare.

I'd suggest adding a short comparison before the Local GPE Routing section.

Something like:
RESTPP supports three query routing modes:

Mode 0 (RoundRobin): Routes queries across GPEs in round-robin order.
Mode 1 (CPULoadAware): Routes queries to the least-loaded GPE based on CPU usage.
Mode 2 (LocalGPE): Routes every query to the GPE running on the same node as the receiving RESTPP instance.

modules/system-management/pages/workload-management.adoc
```diff
@@ -578,3 +578,22 @@ If no GPE satisfies the CPU threshold condition, REST++ falls back to the defaul
...
+
+When `RESTPP.QueryRouting.Mode` is set to `2` (LocalGPE), REST++ forwards each query directly to the GPE co-located on the same node as the receiving RESTPP instance.
+
+This removes the network hop between RESTPP and a remote GPE, reducing query latency for high QPS workloads.
```
Tushar-TG-14: Expand it slightly:

This removes the network hop between RESTPP and a remote GPE, reducing query latency for high QPS workloads. It is most beneficial in multi-node or HA deployments where RESTPP would otherwise forward requests to a remote GPE.

Tushar-TG-14 (changes requested):

modules/reference/pages/configuration-parameters.adoc
```diff
@@ -1075,6 +1075,9 @@ concurrent queries in the delay queue |`20`
 |RESTPP.QueryRouting.Mode | Set the query routing mode (0=RoundRobin, 1=CPULoadAware) | `0`
```
qingGongTG: This config already exists here, here we only add `2=LocalGPE` for this config.
The default value is still 0, not 2.

PR #959, opened 2026-07-09
Update jwt-token.adoc

AyeshaFirdausTG:


modules/user-access/pages/jwt-token.adoc
```diff
@@ -365,6 +365,48 @@ If the user configures this claim and the token does not match with a value in t
...
+==== Security.JWT.UserClaim
+
+Ex: Security.JWT.UserClaim
+
+[source, console]
+----
+$ gadmin config set Security.JWT.UserClaim "upn"
+----
+
+By default, `Security.JWT.UserClaim` is unset (empty), and TigerGraph uses the `sub` claim to resolve the TigerGraph username, as described in *payload*.
```
xun-gao-tigergraph: the default value is 'sub'

xun-gao-tigergraph: when an administrator sets the value to empty: fall back to sub.

xun-gao-tigergraph (approved):
LGTM

